Issue #22, opened 2019-05-08
Need channel numbers and band numbers in eta scan plots


swh76:
Let's add the channel number, band number, and tune file name (or dataset ctime, or whatever) to the eta scan plots.

![image](https://user-images.githubusercontent.com/20823858/57398040-f0bb3a80-7182-11e9-90fa-4df0fff1ae00.png)

eyyoung24:
Closing this issue because channels and bands are not assigned when eta-scans are taken.

PR #23, opened 2019-05-09
Eta plot fix

eyyoung24:
Fix for this request: https://github.com/slaclab/pysmurf/issues/22

Not including channel number because we haven't assigned channels yet. However I added resonator number.


swh76:
Do I have to leave a comment to "review" a pull request?

PR #92, opened 2019-09-12
Packages pysmurf library with setup.py and adds versioneer

jlashner:
One of the comments made at the hack session was that we want to keep track of the pysmurf version used to create output files and plots. Right now there is no way to easily view the pysmurf version at runtime. 

This PR makes pysmurf into a python package by adding the setup.py file, and adds versioneer to keep track of version info, which you can view at runtime by calling `pysmurf.__version__`. 

I haven't been able to test this running actual pysmurf functions, other than checking that I can properly import pysmurf and its subpackages, so maybe you can try it on the system at SLAC to make sure that it works? 

It will also require some changes to the pysmurf-docker container. Instead of adding the pysmurf base to the $PYTHONPATH, you need to run `pip install -e .` from the pysmurf directory. The `-e` makes the package editable, so you can mount dev copies of pysmurf at runtime. I'll also create a PR in the pysmurf-docker repo to do this.


jlashner:
Should this be a version update so we don't have backwards compatibility issues within the same version?

jesusvasquez333:
Correct. After this PR is merged, the following release number should be `v4.0.0`

PR #99, opened 2019-09-25
Fix sphinx build

BrianJKoopman:
The sphinx build doesn't currently work at all, but many of the pysmurf modules are documented in a style not too distant from the [napoleon compatible styles](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html). This cleans up the `docs/` directory a bit and outlines a basic set of `.rst` files to autobuild pages to document the APIs that are already documented.

The styles used in the documentation do seem to be a mix of the Google and Numpy styles, which will cause a slew of warnings when you build. I'd suggest fixing those to clean up the formatting of the built pages, but one style needs to be picked.

To build you'll need pysmurf installed on the system, then in the `docs/` directory run:
```
make html
```

Output is in `_build/html/`. For example, open `_build/html/index.html` after building.


swh76:
@BrianJKoopman Thanks!  If you don't mind (feel free to unassign yourself) I will create an issue assigned to you to add a section to the top level pysmurf README with how to build and look at the docs.

PR #113, opened 2019-10-09
Code and comment cleanup

eyyoung24:
This is a big code clean up. I added comments to a lot of functions. Also removed some deprecated functions. Removed all python2 style string formatting,

Additionally the offline mode was broken due to the validator. Validation is now an optional step (default to True).


pysmurf/tune/smurf_tune.py
```diff
@@ -2628,7 +2453,8 @@ def unset_fixed_flux_ramp_bias(self,acCouple=True):
                 time.sleep(self._cryo_card_relay_wait)
 
 
-    def set_fixed_flux_ramp_bias(self,fractionFullScale,debug=True, do_config=True):
+    def set_fixed_flux_ramp_bias(self,fractionFullScale,debug=True, d
```
swh76: Typo?

pysmurf/tune/smurf_tune.py
```diff
@@ -2214,20 +1991,25 @@ def flux_ramp_check(self, band, reset_rate_khz=4,
...
-    def tracking_setup(self, band, channel=None, reset_rate_khz=4., write_log=False, 
-        make_plot=False, save_plot=True, show_plot=True, nsamp=2**19,
-        lms_freq_hz=None, meas_lms_freq=False, flux_ramp=True, fraction_full_scale=None,
-        lms_enable1=True, lms_enable2=True, lms_enable3=True, lms_gain=7,
-        return_data=True, new_epics_root=None,
+    def tracking_setup(self, band, channel=None, reset_rate_khz=4., 
+        write_log=False, make_plot=False, save_plot=True, show_plot=True, 
+        nsamp=2**19, lms_freq_hz=None, meas_lms_freq=False, flux_ramp=True, 
+        fraction_full_scale=None, lms_enable1=True, lms_enable2=True, 
+        lms_enable3=True, lms_gain=7, return_data=True, new_epics_root=None,
```
swh76: @eyyoung24 Would you mind fixing the reset_rate_khz=4 and lms_gain=7 hardcode defaults here?  lmsGain can be None and if not provided can be taken from the pysmurf config file.  We should do the same for reset_rate_khz - but if you don't want to do that one you can just create a pysmurf issue for adding it to the config file.

pysmurf/tune/smurf_tune.py
```diff
@@ -2470,52 +2264,84 @@ def track_and_check(self, band, channel=None, reset_rate_khz=4.,
         relock=True, tracking_setup=True,
         feedback_start_frac=None, feedback_end_frac=None):
         """
-        This runs tracking setup and check_lock to prune bad channels.
+        This runs tracking setup and check_lock to prune bad channels. This has
```
swh76: @eyyoung24 could you please also fix the lms_gain hardcode here, and add to the reset_rate_khz hardcode issue (or fix).

pysmurf/tune/smurf_tune.py
```diff
@@ -452,41 +456,43 @@ def plot_tune_summary(self, band, eta_scan=False, show_plot=False,
...
             keys = self.freq_resp[band]['resonances'].keys()
             n_keys = len(keys)
+            # If using full band response as input
             if 'full_band_resp' in self.freq_resp[band]:
                 freq = self.freq_resp[band]['full_band_resp']['freq']
                 resp = self.freq_resp[band]['full_band_resp']['resp']
                 for k in keys:
                     r = self.freq_resp[band]['resonances'][k]
-                    width = .300 # 300 kHz
+                    width = .300 # in MHz. So this = 300 kHz
```
swh76: @eyyoung24 Fix hardcode!

pysmurf/tune/smurf_tune.py
```diff
@@ -689,27 +696,29 @@ def find_peak(self, freq, resp, rolling_med=True, window=5000,
...
 
+        # Break apart the data
         angle = np.unwrap(np.angle(resp))
         x = np.arange(len(angle))
         p1 = np.poly1d(np.polyfit(x, angle, 1))
         angle -= p1(x)
-        grad = np.convolve(angle, np.repeat([1,-1], 8),
-                           mode='same')
-        #grad = np.ediff1d(angle, to_end=[np.nan])
+        grad = np.convolve(angle, np.repeat([1,-1], 8), mode='same')
```
swh76: @eyyoung24 should the 8 hardcode here also be made a function argument?  Too lazy to look up np.convolve.

pysmurf/tune/smurf_tune.py
swh76: Oh I see you caught this one already in the last commit in this group.  Resolved.

pysmurf/tune/smurf_tune.py
eyyoung24: @swh76 Can you point me to where the lms_gain=None check is? I'm not obviously finding it. Want to make sure it's doing the right thing.

swh76: @eyyoung24 oh sorry, I just meant to implement code for lms_gain in that (and other) functions like this ; 
https://github.com/slaclab/pysmurf/blob/b397b6d7382fff7794a367225ed80b6a319381f6/pysmurf/util/smurf_util.py#L2462
where if none is explicitly provided, it will default to the value in the pysmurf cfg file.

Issue #119, opened 2019-10-21
Function for requesting filter params


eyyoung24:
SO wants to be able to request the filter the params via a pysmurf call.

PR #120, opened 2019-10-21
Adding function to get downsample filter parameters

eyyoung24:
This resolves https://github.com/slaclab/pysmurf/issues/119

### DIFF
--- a/pysmurf/util/smurf_util.py
+++ b/pysmurf/util/smurf_util.py
@@ -2590,7 +2590,66 @@ class SmurfUtilMixin(SmurfBase):
 
         return ret
 
+    
+    def get_smurf_to_mce_file(self):
+        """
+        Reads in the most recent smurf_to_mce file and returns
+        everything in it as a dict.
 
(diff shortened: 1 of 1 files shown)


mhasself:
I think there's a small bug here -- If I'm reading [smurf_processor](https://github.com/slaclab/smurf-processor/blob/bfe00ed6f4e1ed9ee629f25a68a79fa1e3076eda/src/smurf_processor.cpp#L995-L1002) correctly, then `filter_order`:
- when non-negative, indicates that `a` and `b` have up to `filter_order + 1` non-zero values.
- can be -1, which will cause np.zeros to fail (the filter behavior in this case isn't totally clear to me)

swh76:
Re: (at least the intended) behavior for negative filter_order, see Joe's documentation here:
https://confluence.slac.stanford.edu/display/SMuRF/SMURF2MCE

> Filter_order
> 
> This selects what order low pass filter will be used. -1 selects the old rectangular integrator.

PR #140, opened 2019-10-31
Process data files with dynamic data frame sizes. 

jesusvasquez333:
https://jira.slac.stanford.edu/browse/ESCRYODET-425


eyyoung24:
the pyrogue session crashes if SmurfProcessor:ChannelMapper:NumChannels > PayloadSize. I can write a check in pysmurf, but this feels like it should be in pyrogue.

jesusvasquez333:
Thanks for catching that bug @eyyoung24. I will take a look. You are right, that should be allowed, so I don't think you need to add a check for that

PR #152, opened 2019-11-05
Check if the EPICS features were requested before adding them

jesusvasquez333:
Check if the EPICS server is requested by the user before starting it. Also, check if the other features related to it (stream data to PV, and PV dump file) were requested as well.

https://jira.slac.stanford.edu/browse/ESCRYODET-441


slacrherbst:
Are there ever any cases where we don't start epics?

PR #156, opened 2019-11-07
Smurf reader class

slacrherbst:
Here I have created a SmurfStreamReader class which can be used to process both new and old data formats, with meta data support. This also supports reading data broken into multiple files:

Example usage:

`````
        with SmurfStreamReader(fileList,isRogue=True, metaEnable=True]) as fr:

            for header,data in fr.records():

                 # Process data as int32 numpy array

`````

### DIFF
--- a/python/pysmurf/client/util/SmurfFileReader.py
+++ b/python/pysmurf/client/util/SmurfFileReader.py
@@ -73,7 +73,7 @@ RogueHeader = namedtuple( 'RogueHeader',
 
 
 
-class SmurfFileReader(object):
+class SmurfStreamReader(object):
 
     def __init__(self, files, *, isRogue=True, metaEnable=False, chanCount=None):
         self._isRogue    = isRogue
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: python/pysmurf/client/util/SmurfFileReader.py [hunk 1]
```diff
@@ -0,0 +1,254 @@
...
+                if self._currFile.tell() == self._fileSize:
+                    return False
+
+                # Not enough data left in the file
+                if (self._fileSize - self._currFile.tell()) < RogueHeaderSize:
+                    printf(f"Waring: File under run reading {self._currFName}")
+                    return False
+
+                # Read in Rogue header data
+                rogueHeader = RogueHeader._make(struct.Struct(RogueHeaderPack).unpack(self._currFile.read(RogueHeaderSize)))
```
jesusvasquez333: @slacrherbst Did you verify if this line unpack the rogue header correctly? I tried doing it like this, but it was not unpacking the second 32-bit value correctly. So, I had to unpack the rogue header like this: 

https://github.com/slaclab/pysmurf/blob/355067995c91640fe15e783d2a62b81794dbb82c/python/pysmurf/client/util/smurf_util.py#L841-L867

slacrherbst: I am working on a core rogue version of this. I will test it there and update this PR if required.

slacrherbst: The rogue header unpack is working for me in testing.

PR #162, opened 2019-11-11
Create seperate Rogue Version of Publisher

slacrherbst:
I suspect the API will end up being a bit different for the Rogue version of the publisher, so I have copied the publisher code into the Rogue interface.

### DIFF
--- a/python/pysmurf/core/utilities/_SmurfPublisher.py
+++ b/python/pysmurf/core/utilities/_SmurfPublisher.py
@@ -95,7 +95,7 @@ class SmurfPublisher(object):
         # Process root looking for variables part of the publish group
         for v in root.variableList:
             if v.inGroup('publish'):
-                v.addVarListener(self._varListen)
+                v.addListener(self._varListen)
 
         # Issue the start message.
         self.log_start()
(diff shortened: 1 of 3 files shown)


python/pysmurf/core/utilities/_SmurfPublisher.py
```diff
@@ -17,27 +17,179 @@
...
+                             (self.__class__, backend))
+            self._backend = self._backend_null
+
+        # ###########################
+        # Rogue Specific Code
+        # ###########################
         # Process root looking for variables part of the publish group
         for v in root.variableList:
             if v.inGroup('publish'):
                 v.addVarListener(self._varListen)
```
jesusvasquez333: Is this a typo? I think it should be
```
v.addListener(self._varListen)
```

python/pysmurf/core/utilities/_SmurfPublisher.py
```diff
@@ -17,27 +17,179 @@
...
+            'format': format,
+            'timestamp': timestamp,
+            'plot': plot,
+            'pysmurf_version': 'replace_with_version'
+        }
+
+        return self.publish(file_data, 'data_file')
+
     def _varListen(self, path, varVal):
         self.pub.publish(data=f'{path}={varVal.value}',msgType='general')
```
jesusvasquez333: `self` doesn't have a `pub` method. Should this call `self.publish` instead? 
Also, the `publish` method doesn't have a `msgType` argument, but instead a `msgtype`. 

So, I think this line should be:
```
self.publish(data=f'{path}={varVal.value}',msgtype='general')
```

slacrherbst: Ok these have all been addressed.

PR #179, opened 2019-11-14
Fix automatic documentation builds

BrianJKoopman:
Builds on readthedocs were failing silently due to maptlotlib not being in `requirements.txt`. This just adds it, fixing the documentation builds.


swh76:
Thanks!

PR #219, opened 2020-02-01
Accelerated processing of TES bytes

slacrherbst:
TES bytes are now processed when the header is read, using a faster approach than before:

Two changes to the header class:

To obtain external_time value: val = header.external_time
To obtain TES bias values: val = header.tesBias[n]


jesusvasquez333:
Changing this PR to `master` as we plan to remove `pre-release`

Issue #95, opened 2019-09-18
Add Dan's RF initialization fix to pysmurf setup and test


swh76:
Dan may have found fix for the system sometimes coming up with a (or a few) 500MHz bands' RF DACs not configured properly (more details on the JIRA ticket here - https://jira.slac.stanford.edu/browse/ESCRYODET-282).  Below is the fix @jmdewart implemented in the pyrogue tarball ; should port into pysmurf setup at the end, and test to see if this solves the problem.

```python
# Special DAC Init procedure (pyrogue implementation)
            for dac in dacDevices: 
                dac.EnableTx.set(0x0)
                dac.InitJesd.set(0x1)
                dac.JesdRstN.set(0x0)
                dac.JesdRstN.set(0x1)
                dac.InitJesd.set(0x0)
                dac.EnableTx.set(0x1)
            if len(dacDevices) > 0:
                for lmk in lmkDevices:
                    lmk.PwrUpSysRef()
```

swh76:
After some testing, RF initialization is still failing after boot on at least one band 20% of the time with this fix (with the above pyrogue implementation - haven't moved this to pysmurf yet).  Will leave the issue open because the underlying problem remains, but the fix suggested above is not enough to completely resolve this issue.

swh76:
Turns out this is probably due to issue with JESD deterministic latency (see ESCRYODET-401), solved in firmware version 6d670ac and later.  After fixing the JESD deterministic latency, @jmdewart was able to boot and successfully initialize the RF DACs 400x in a row.

PR #229, opened 2020-02-20
Issue95

swh76:
This is a very badly named branch.  This branch was developed in RF lab and morphed from an implementation of https://github.com/slaclab/pysmurf/issues/95 into all the scripts and enhancements needed to do hardware/thermal testing.


eyyoung24:
@swh76 Change save_data and then should be fine to merge.

PR #240, opened 2020-02-26
Fixing bug with loading mask

eyyoung24:
Small change that parses the mask file properly.


eyyoung24:
see https://github.com/slaclab/pysmurf/pull/240

Issue #244, opened 2020-02-26
make plots for plot_tune_summary in a separate thread


eyyoung24:
As a test case for multiprocessing plots, make the tuning summary plots in a separate thread. Something like this: https://stackoverflow.com/questions/51409295/multiprocess-plotting-in-matplotlib

ruck314:
Status check on this 5-year-old enhancement: still not implemented in `main` as of 5f0b34eb. `plot_tune_summary` (`python/pysmurf/client/tune/smurf_tune.py:435-578`) is synchronous, and there is no `threading` / `multiprocessing` use in the tune module.

Worth noting: the original author opened #245 to implement this and closed it without merging. Their own comments on that PR explain why it stalled — `multiprocessing.Process` with `fork` could not safely share the parent's rogue/EPICS connections, so some PV reads worked from the child and others didn't. They worked around it for `band_center` / `ref_phase_delay*` by hoisting reads into the parent, but `plot_eta_fit` does its own register reads, so the approach was incomplete.

A correct implementation would need to (a) pre-fetch all PV data in the parent, (b) refactor `plot_eta_fit` to be picklable / register-free in the worker, and (c) either use `multiprocessing` with the `spawn` start method + `matplotlib.use('Agg')` in the child, or use matplotlib's OO API (`Figure(...) + FigureCanvasAgg`) under a `threading.Lock`. The issue itself is described as a "test case" rather than user-facing pain, no one has revisited it in five years, and `eta_scan=True` is opt-in for the slow path. Recommend closing as won't-fix unless someone has a current need.

PR #245, opened 2020-02-26
Spawning new thread for making tune summary plots.

eyyoung24:
https://github.com/slaclab/pysmurf/issues/244

### DIFF
--- a/python/pysmurf/client/tune/smurf_tune.py
+++ b/python/pysmurf/client/tune/smurf_tune.py
@@ -396,7 +396,6 @@ class SmurfTuneMixin(SmurfBase):
         self.log('Making plots in a separate thread. ' +
                  'Suppressing output.')
 
-        
         import multiprocessing as mp
         
         # Create process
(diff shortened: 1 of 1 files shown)


python/pysmurf/client/tune/smurf_tune.py
```diff
@@ -379,8 +379,39 @@ def tune_band_serial(self, band, n_samples=2**19,
...
+        """
+        Opens a new thread to make tune_summary plots.  This 
+        uses the multiprocessing module.
+
+        Args:
+        -----
+        band (int) : The band to make the plots in.
+        """
+        # Get some useful values that are repeatedly plotted
+        ref_phase_delay = self.get_ref_phase_delay(band)
```
eyyoung24: @slacrherbst I can't seem to get these variables in my new thread, so I have to get them in my main thread. Other PVs I can request. Any idea why some you can, and others you cannot?

eyyoung24:
Does this seem like a reasonable way to spawn plotting in a new thread. I worry that you could in principle start too many threads though if you're not careful.

PR #250, opened 2020-02-26
Pton pre-release and issue 249 fix

swh76:
Updates Princeton configuration scripts in master to what's being used for 2xLB testing on 2/26/2020, and adds fix for issue https://github.com/slaclab/pysmurf/issues/249.

Issue #249, opened 2020-02-26
'smurf_chans' referenced before assignment bug when running take_noise_psd


swh76:
In master branch (and pre-release), Heather & Max report this error when running `take_noise_psd` - "local variable 'smurf_chans' referenced before assignment".  This is caused by a pysmurf indent bug here:

https://github.com/slaclab/pysmurf/blob/cf467bca4b6157c9cd7986eaa72ca1e2b925eb63/python/pysmurf/client/util/smurf_util.py#L736

I changed that block of code to 

```python
# Optionally write PyRogue configuration
            if write_config:
                config_filename=os.path.join(self.output_dir, timestamp+'.yml')
                if write_log:
                    self.log('Writing PyRogue configuration to file : '+
                         f'{config_filename}', self.LOG_USER)
                self.write_config(config_filename)

                # short wait
                time.sleep(5.)
            if write_log:
                self.log(f'Writing to file : {data_filename}',
                         self.LOG_USER)

            smurf_chans = {}
            for b in bands:
                smurf_chans[b] = self.which_on(b)
            output_mask = self.make_channel_mask(bands, smurf_chans)
            self.set_channel_mask(output_mask)
```
which I think resolves the error.  Need to create a branch with this fix to merge into master (because Princeton's still on pre-release branch, I can't merge the fix directly from there).  I'll try to do that right now.

eyyoung24:
Fixed here: https://github.com/slaclab/pysmurf/pull/253

PR #251, opened 2020-02-26
Pton pre-release and issue 249 fix #250

swh76:
Ok merging this to pre-release, since that is probably our workflow (just attempted to PR to master).  Updates Princeton configuration scripts in master to what's being used for 2xLB testing on 2/26/2020, and adds fix for issue #249.


jesusvasquez333:
Changing this PR back to master, as we are going to remove the `pre-release` branch and use only master.

PR #257, opened 2020-02-27
Issue254 - function for gettting all frequencies in a band

eyyoung24:
This commit also has a lot of shawn's scratch updates


eyyoung24:
This also spits out a <timestamp>_freq.txt which is paired with <timestamp>_mask.txt when writing data to disk. The function read_stream_data() will also optionally return this mask. The default is to NOT return the frequency mask to preserve backwards compatability.

PR #285, opened 2020-03-07
Detect type of carrier and enable the EM22xx device if it is C03+

jesusvasquez333:
This PR adds the ability to the pysmurf-server startup script to autodetect the type of carrier via IPMI, and if the carrier is Gen2 and version >= C03, enable the EM22xx device. 

In order for this to be fully operational, we need:
- This [PR#31](https://github.com/slaclab/cryo-det/pull/31) in the `cryo-det` repository.
- I found issue with the EM22xx device definition, so the JIRA ticket [ESCRYODET-614](https://jira.slac.stanford.edu/browse/ESCRYODET-614) needs to be solved, and then `cryo-det` needs to bring that fix.

After those 2 steps are completed, we will need to generated a new ZIP file from the `cryo-det` repository. 

To avoid problems until we get a new ZIP file, in this PR the EM22xx device will always be disabled; and it is also backwards compatible with current ZIP files.. So, for now I'm just adding all the pieces we need to support this feature. Once we are ready we can enable this feature by only changing these 2 lines:
https://github.com/slaclab/pysmurf/blob/d2b570b991a5b478d8cfaa7eaaf10286474640ff/server_scripts/cmb_eth.py#L61
https://github.com/slaclab/pysmurf/blob/d2b570b991a5b478d8cfaa7eaaf10286474640ff/server_scripts/cmb_pcie.py#L57

For details see: https://jira.slac.stanford.edu/browse/ESCRYODET-542


jesusvasquez333:
I took a look at the different repositories and I didn't see any IntelEnpirion device there. I found it in some Rogue3 tarballs in smurf-srv11. So, my guess is that it was locally added to the tarballs, but never push officially to any repository.

PR #295, opened 2020-03-10
Add argument to publish streamed data file 

### LINKED ISSUE #289: Add argument to publish streamed data file
An argument should be added to stream_data_off (or maybe stream_data_on makes more sense?) that determines if the streamed data file should be registered via the publisher.

jlashner:
This issue resolves #289 by adding the `register_file` argument to `stream_data_off`. When specified, the `.dat` file created during the streaming session will be registered with the publisher. The idea is that we want to archive smurf-data for most pysmurf functions including tuning and iv, but don't want to archive data when we are streaming for arbitrary lengths of time.

### DIFF
--- a/python/pysmurf/client/util/smurf_util.py
+++ b/python/pysmurf/client/util/smurf_util.py
@@ -849,14 +849,13 @@ class SmurfUtilMixin(SmurfBase):
                 If true, the stream data file will be registered through the
                 publisher.
         """
-        if register_file:
-            datafile = self.get_data_file_name().tostring().decode()
-
         self.close_data_file(write_log=write_log)
 
         if register_file:
-            self.log("Registering File {}".format(datafile))
-            self.pub.register_file(datafile, 'data', format='dat')
+            datafile = self.get_data_file_name().tostring().decode()
+            if datafile:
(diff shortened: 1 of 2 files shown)


eyyoung24:
Do you know the new flake8 requirements? I can talk you through it if you don't @jlashner

jlashner:
Alright, fixed!

python/pysmurf/client/util/smurf_util.py
```diff
@@ -835,11 +838,26 @@ def stream_data_on(self, write_config=False, data_filename=None,
...
         Turns off streaming data.
+
+        Args:
+            write_log (bool):
+                Whether to log the CA commands or not. Default False
+            register_file (bool):
+                If true, the stream data file will be registered through the
+                publisher.
         """
+        if register_file:
```
jesusvasquez333: I think can refactor this section of the code to be something like:
```
self.close_data_file(write_log=write_log)

if register_file:
    datafile = self.get_data_file_name().tostring().decode()
    self.log("Registering File {}".format(datafile))
    self.pub.register_file(datafile, 'data', format='dat')
```
`get_data_file_name()` will return the name of the data file, even after the file is closed.

PR #302, opened 2020-03-11
Makes shawnhammer shutdown smurf-streamer dockers

jlashner:
This change makes it so that shawn-hammer kills possibly conflicting dockers that were started by jackhammer.


scratch/shawn/scripts/shawnhammer.sh
```diff
@@ -129,6 +129,13 @@ if [ "$?" = "1" ]; then
     docker rm -f $(docker ps -q -f name=pysmurf | awk '{print $1}')
 fi
 
+# stop all smurf-streamer dockers
+matching_dockers smurf-streamer
+if [ "$?" = "1" ]; then
+    echo "-> Stopping all running stable pysmurf dockers."
```
swh76: Please differentiate the echo statement from the preceeding.  Maybe "Stopping all running smurf-streamer dockers"?

jlashner:
Good catch. Fixed.

PR #308, opened 2020-03-13
Issue306 - fixing bias bump - introduce estimate_bias_voltage function that estimates where to bias using an input IV curve

eyyoung24:


eyyoung24:
This uses the header for TES bias instead of just guessing.

python/pysmurf/client/debug/smurf_iv.py
```diff
@@ -1171,4 +1172,8 @@ def estimate_bias_voltage(self, iv_file, target_r_frac=.5,
 
             if not show_plot:
                 plt.close(fig)
-        return target_bias_voltage
+
+        targets = np.zeros(8)
```
swh76: We only had 8x TES bias lines at Pole, but need to support all 16.

eyyoung24: Makes sense. But isn't it 15 bias groups?

swh76: Yea it's complicated, and the best way to handle this isn't obvious.  Some hopefully helpful information:

1) Here's the signals on the cable from the RTM to the cryostat card:
image
the DAC pairs that can be used as bipolar TES biases are the DAC nets.  There are 32 of those total, so a maximum of 16 possible bipolar TES biases from the RTM total.  

2) 15 is probably in your head because the current SO cryostat card ([PC-248-103-02-C02](https://confluence.slac.stanford.edu/display/AIRTRACK/PC-248-103-02-C02)) has jumpers on it which can optionally repurpose DAC31/32 as gate voltages for a 2nd HEMT+50K amplifier (meant to provide for using SMuRF with 1000x MUX factor):
image
To further complicate things, the SO cryostat card only actually uses 12 DAC pairs from the RTM as TES biases.  But for instance for BA HF we may end up using all 16 DAC pairs from the RTM as TES biases.

3) We had a lot of hardcode 8s for TES biases and no way of designating RTM DACs as being used for biasing amplifiers instead of TESs but I attempted to make our code more general in this PR;https://github.com/slaclab/pysmurf/pull/107.  I attempted to make it flexible and driven from parameters (not hard-codes) but I don't claim it's the best approach.

jesusvasquez333: @eyyoung24 Maybe, instead of hardcoding `8` you should do:
```
targets = np.zeros(len(bias_group))
```

Or do you want to return a fix size array?

eyyoung24: changed to:
`targets = np.zeros(self._n_bias_groups)`

python/pysmurf/client/debug/smurf_iv.py
```diff
@@ -1086,9 +1086,10 @@ def estimate_bias_voltage(self, iv_file, target_r_frac=.5,
         save_plot (bool) : Whether to save the plot. Default True.
         """
```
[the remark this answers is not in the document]
swh76: Ed fixed with commit 22b085f.

Issue #311, opened 2020-03-13
Fix RF amplifier drain current resistor hard-codes


swh76:
These should be made into optional cfg file entries whose default values if they aren't present in a user's cfg file is the currently hard-coded values.  No hard codes (pretty sure these were my fault, so assigning to myself).

https://github.com/slaclab/pysmurf/blob/99e5d2da93521bf9e02e8b4ab3544f5a2ba252c2/python/pysmurf/client/util/smurf_util.py#L2561

https://github.com/slaclab/pysmurf/blob/99e5d2da93521bf9e02e8b4ab3544f5a2ba252c2/python/pysmurf/client/util/smurf_util.py#L2574

swh76:
Closed with PR https://github.com/slaclab/pysmurf/pull/313.

PR #313, opened 2020-03-13
Issue311

swh76:
Addresses issue https://github.com/slaclab/pysmurf/issues/311.  Somehow some of Ed's commits got in here (I thought I reset to master but I apparently failed).  @eyyoung24 let me know if I need to remove them.

### DIFF
--- a/python/pysmurf/client/util/smurf_util.py
+++ b/python/pysmurf/client/util/smurf_util.py
@@ -3060,9 +3060,10 @@ class SmurfUtilMixin(SmurfBase):
 
 
     def bias_bump(self, bias_group, wait_time=.5, step_size=.001, duration=5,
-                  fs=180., start_bias=None, make_plot=False, skip_samp_start=10,
+                  start_bias=None, make_plot=False, skip_samp_start=10,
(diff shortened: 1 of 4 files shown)


python/pysmurf/client/util/smurf_util.py
```diff
@@ -2551,29 +2551,73 @@ def get_amplifier_biases(self, write_log=True):
...
         Returns:
         --------
-        cur (float): The drain current in mA
+        cur (float): 50K amplifier drain current in mA
         """
 
-        # on cryostat card
-        fiftyK_amp_Vd_series_resistor=10 #Ohm
+        # assumes circuit topology on rev C2 cryostat card
+        # (PC-248-103-02-C02, sheet 3)
```
eyyoung24: I assume this is just a reference to a drawing? This is probably just useful internally? That's fine either way, just curious.

swh76: Yea, the rev C2 cryostat card schematic.  These functions basically just implement the circuit from that sheet.  Agreed that it's awkward to refer to a schematic that only we can access, but that's the only source material.

PR #315, opened 2020-03-15
Fix shawnhammer smurf server detection

swh76:
shawnhammer now detects when the pysmurf server is up by waiting until it gets a valid timestamp from the server (smurf_server_s#:AMCc:LocalTime) and that timestamp is incrementing in both serial and parallel bootup modes.  Previously shawnhammer detected server up by looking for a print statement in the rogue docker, which has already broken a couple of times when that print statement was changed.  The print statement shawnhammer used to parse was the "GUI is up" print statement, so now shawnhammer can detect the server is up even if the GUI is not enabled, and can setup faster, since it seems like the server often comes up before the GUI does.  Also fixes an issue with attaching at the end of pysmurf setup ; the function in shawnhammer that detects when pysmurf is done setting up (by looking for the "Done with setup" print statement at the end of pysmurf setup) wasn't grabbing enough lines of the tmux buffer to catch the statement for some versions of pysmurf where extra printout after setup in the python terminal was pushing the "Done with setup" statement out of the buffer being parsed for that statement.

### DIFF
--- a/cfg_files/stanford/fp30_smurf_startup.cfg
+++ b/cfg_files/stanford/fp30_smurf_startup.cfg
@@ -2,12 +2,8 @@ shelfmanager=shm-smrf-sp01
 
 set_crate_fans_to_full=true
 # COMTEL max fan level is 100, ASIS is 15, ELMA is 15
-## COMTEL in RF lab
+## We're using an ELMA crate on campus right now.
 max_fan_level=10
-## ELMA in RF lab
-#max_fan_level=15
-## ASIS in RF lab
-#max_fan_level=15
 
 attach_at_end=true
 screenshot_signal_analyzer=false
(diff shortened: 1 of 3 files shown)


cfg_files/stanford/fp30_smurf_startup.cfg
```diff
@@ -3,7 +3,7 @@ shelfmanager=shm-smrf-sp01
 set_crate_fans_to_full=true
 # COMTEL max fan level is 100, ASIS is 15, ELMA is 15
 ## COMTEL in RF lab
-max_fan_level=15
+max_fan_level=10
```
eyyoung24: @swh76 The comment says the max fan leve is 100, but the var is 10

swh76: Max for ELMA is 15 which is what we have on campus, this is correct albeit not at all obvious.  JIRA exists for auto detecting crate manufacturer to make this less arbitrary.

eyyoung24: Do we want to change the comment? I'll approve the PR either way.

PR #316, opened 2020-03-15
Adds default server_port when IP Address is not defined

jlashner:
This PR adds a default server_port in the process_args function. This is to support running in emulation mode when neither the pcie_rssi_lane or the ip address is defined.


jesusvasquez333:
The server port argument was not implemented for the emulator. So, I aded that new feature, and now the server port argument is propagated from the top server script all the way to the Common root.

PR #317, opened 2020-03-16
Publisher actions

### LINKED ISSUE #56: Need listener example for SO publisher
We need a simple example for how to listen to the SO publisher to help debug.  IE, say I want to make sure that a function I implement is properly writing all of the paths to its output data files to the publisher.

jlashner:
This PR is to solve the OCS problem for how we should be storing archived pysmurf data. From the workshop, we decided that a good system would be to group together all files for a given action. For instance, if you call S.find_freq, you might naturally want all outputs and plots to be grouped together under something like `smurf_dir/find_freq_1584335008/outputs` and `smurf_dir/find_freq_1584335008/plots`.

The way this worked before was that each `register_file` call was passed a "file type" such as "mask", which was used to group files. This didn't really work since functions like `find_freq` would call other functions, and so even though you only issued one command they would multiple groups.

The `set_action` decorator solves this by making the top-level function call set the action and action timestamp, which is then used in all nested function calls. When the top-level function returns, or if an error is thrown, the action and timestamp are set to None.  

You can set the action by decorating a smurf function with `@set_action('find_freq')`, or you can use `@set_action()` which will set the action string to be the function name. 

I added the set_action decorator to a bunch of functions (probably more than we actually need). I also added the script `scratch/jack/pub_listener.py` which will listen to the publisher and resolves #56 . 

This will probably replace the "file type" system, but I'm keeping it in temporarily for backwards compatibility.

### DIFF
--- a/python/pysmurf/client/debug/smurf_iv.py
+++ b/python/pysmurf/client/debug/smurf_iv.py
@@ -20,9 +20,11 @@ import os
 import matplotlib.pyplot as plt
 from matplotlib.gridspec import GridSpec
 import matplotlib.colors as Colors
+from pysmurf.client.util.pub import set_action
 
 class SmurfIVMixin(SmurfBase):
 
+    @set_action()
     def slow_iv_all(self, bias_groups=None, wait_time=.1, bias=None,
                     bias_high=1.5, bias_low=0, bias_step=.005,
                     show_plot=False, overbias_wait=2., cool_wait=30,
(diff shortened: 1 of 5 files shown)


eyyoung24:
Have you tested all these function calls and make sure it doesn't break anything?

jlashner:
I've tested `stream_data_on/off` and `take_stream_data` with the emulator which verifies that the set_actions decorator works correctly. I haven't been able to test all of the functions (like find_freq/tracking_setup) since I don't have a carrier to run it on. Is there one available at SLAC?

jlashner:
Right, you can also set the action name when calling a function by including the `pub_action` keyword. This is added to all functions with the `@set_action` decorator and doesn't mess with the kwargs actually sent to the pysmurf function.

I agree we should test `tracking_setup` and `find_freqs` on a carrier, but the `take_stream_data` function also uses nested function calls and that has been tested.

jlashner:
Alright sorry this took so long, but tested find_freqs and tracking_setup on the UCSD system and can confidently say that it doesn't break. Are we ok if I rebase/merge this?

jlashner:
Alright! Rebased and am just waiting for approval.

Issue #321, opened 2020-03-18
Enforce PRs to have at least one of the labels `bug`, `enhancement`, or `interface-change`


jesusvasquez333:
Setup a GitHub workflow to enforce the use of one of these mandatory labels:
- `bug`
- `enhancement`
- `interface-change`

before merging a PR. 

These labels will be used to autogenerate release notes

jesusvasquez333:
I did some research and some testing regarding this feature and unfortunately I don't think there is a convenient way to implement this. 

One option I tested was to use travis to run a python script too check if the PR has one of the mandatory labels. In travis a stage to run the script will be trigger by the condition `if: type = pull_request` to make sure the test runs only PRs. However, the issue is that travis is only triggered when the PR is created, and with every new push to the PR. This means, for example, that is one opens a PR without adding a label initially, then the travis test will fail, but adding a label later won't trigger travis again to update the result of the test; only doing a new push will do it, which is not what we want.

Another option is to use a Github workflow. Theses have the advantage that they can be trigger both when a PR is crated as well as when a label is added to the PR. Also, I found a GitHub action in the marketplace called [enforce-pr-label](https://github.com/marketplace/actions/enforce-pr-label) which does exactly what we need, verify that a PR has a proper label. However, during my tests I found out that the two trigger conditions (opening a PR, and adding a label to a PR) will create 2 independent checks in GitHub. This means that, if for example one opens a PR without a label, then a GitHub check will fail; then adding a label will start a second Github check that will succeed, but the first check will remain in the failed state, and the PR's merge will be block. 

Additionally to theses issues, there is a more fundamental issue: even if we enforce a PR to have labels before they re merge, labels on a merged PR can be change anyways. So, it is unlikely but it is possible that an user remove a label of a merged PR. Then, at the time to generate the release notes, the PR could be missing the label we need.

slacrherbst:
Ok. We will just be sure to properly catch the missing label condition in the release notes generation.

jesusvasquez333:
I ended up writing a new GitHub action to work around this issue: https://github.com/marketplace/actions/verify-pull-request-labels. 

Instead of relying on the Check status, it uses pull request reviews request instead. I will open a PR now.

This will make sure a PR has at least one of the mandatory labels. But the fact that labels can be added/deleted on a PR after it has been merged remains. But at least that is less likely to happen.

PR #325, opened 2020-03-20
Verify that all PRs are labeled with `bug`, `enhancement`, or `interface-change`

jesusvasquez333:
This PR solved #321.

It adds the GitHub actions [verify-pull-request-labels](https://github.com/marketplace/actions/verify-pull-request-labels) to verify that all PRs have at least one of these labels before they can be merged: `bug`, `enhancement`, `interface-change`.

This action works around the issue I mentioned in #321 regarding checks trigger with different conditions, and instead works in the following way:
- If a PR does not have one of the valid labels, then the Github check will succeed, but it will create a pull request review, requesting for changes. This will prevent the PR to be merge until the review is addressed. 
- If a PR have at least one valid label, then the Github check will also succeed, and it will create a pull request review, approving it. This will clear the review requesting for changes from a previous failed check. 
- The action will be triggered when a pull request is open, updated, and when labels are added/removed. 

As the Github Check alway succeed, we will not have the problem I mentioned in #321. Instead, the PR will be block using normal pull request workflow.

An example scenario is:
- A PR is open, without one of the mandatory labels. The action runs and creates a PR review, requesting for changes. This will prevent the PR to be merged.
- One of the mandatory labels is added to the PR. The action will be triggered again, and will create a new PR review, approving it. This will dismiss the previous request for changes. 
- If the label is removed, it will trigger the action again, and step 1 will happen again.
- If the PR labeled correctly before opening the PR, then the first action will just crate a PR review approving it.

So, in order for this new feature to work, we will need to increase the number of `Required approving reviewers` from `1` to `2`: one review will always come from this action, and the other review will be the one needed from a real reviewer.

This PR itself will be an example of how this feature works. I'm opening the PR without any label, in order to let the action to crate the PR review, request changes. Then I will add a correct label to fix it.

### DIFF
--- a/.github/workflows/verify-pr-labels.yml
+++ b/.github/workflows/verify-pr-labels.yml
@@ -1,0 +1,21 @@
+# This workflow will verify that all pull requests have at least
+# one of these labels: 'bug', 'enhancement', 'interface-change'
+# before they can be merged
+
+name: verify-pr-label-action
+
+on:
+  pull_request:
+   types: [opened, labeled, unlabeled, synchronize]
+
+jobs:
+  verify_pr_labels:
+    runs-on: ubuntu-latest
+    name: Verify that the PR has a valid label
+    steps:
+    - name: Verify Pull Request Labels
+      uses: jesusvasquez333/verify-pr-label-action@v1.1.0
+      id: verify-pr-label
+      with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+          valid-labels: 'bug, enhancement, interface-change'


jesusvasquez333:
At this point, the PR can not be merged, as there is a request for changes, indicating that a mandatory label is missing. 
I will add the label next.

jesusvasquez333:
Now that I added a correct label, the PR can be merge normally. Increasing the `Required approving reviewers` to `2`, one more review will be needed, from one of us.

eyyoung24:
Do we need to require 2 approvals to pass? github-actions[bot] will always pass (once we have the right tags). But I think we want to also have a human check it, right?

eyyoung24:
Annotation 2020-03-20 104442

eyyoung24:
You can see in the image above that we can already merge it even though no human has approved it.

jesusvasquez333:
@eyyoung24 Yes, you are correct. I explained that in the description of this PR:

> So, in order for this new feature to work, we will need to increase the number of `Required approving reviewers` from `1` to `2`: one review will always come from this action, and the other review will be the one needed from a real reviewer.

PR #327, opened 2020-03-20
Adding and cleaning up comments - also fixes bug in take_noise_psd plot label

eyyoung24:


eyyoung24:
Also deleted eta_reestimator function because it is not used.

Issue #123, opened 2019-10-22
Overhaul amplitudeScale and UC/DC attenauator set/get functions


swh76:
Some ideas for making setting/getting tone power levels easier to understand:

- It's not clear enough to users that amplitudeScale should be set to 12, or what that even means.  Comments about this should be added liberally throughout the docstrings.
- We should add an required manual override with a warning so that users stay cognizant of the possible issues that could arise from running with amplitudeScale< or >12 (in particular, running with amplitudeScale>12 can be problematic if the DAC starts railing).
- Dan suggests setting attenuator and power levels in dB (relative to 12 for amplitudeScale, and relative to no attenuation for the DC and UC attenuators). 

Should also address https://github.com/slaclab/pysmurf/issues/114 at the same time.

eyyoung24:
@swh76 Is it .5 db/step?

PR #333, opened 2020-03-21
Extract issue references from PR body

slacrherbst:
This change will scan the pull request body for the #123 pattern in an attempt to find references to issues.

### DIFF
--- a/releaseGen.py
+++ b/releaseGen.py
@@ -88,15 +88,24 @@ for line in loginfo.splitlines():
             else:
                 entry['Labels'] += ', ' + lbl.name.lower()
 
-        # Attempt to locate any issues mentioned in the body
+        # Attempt to locate any issues mentioned in the body and comments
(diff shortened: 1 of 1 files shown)


jesusvasquez333:
Issues can also be mentioned in comments in the PR, so I modify the script to look in comments as  well. 
Also, I modified to avoid reporting the same issue number multiple times, if the same issue is mentioned more than once in the PR.

Unfortunately, an issue can also be manually linked to a PR, without mentioning in the description to comments. But I didn't find a way to get those. So, those won't be reported in the release notes.

slacrherbst:
It appears that the Issue list can get reset with each comment.

slacrherbst:
Nevermind I misread.

slacrherbst:
Also it appears that issue and PR numbers are generated from the same pool. At least it appears that way.

PR #354, opened 2020-03-25
Fixes the noise_vs_tone function. The plotting routine had a bug in it which would cause it to crash.

### LINKED ISSUE #353: `noise_vs_tone` plotting broken
When `analyze=True`, it attempts to make a plot which fails.

eyyoung24:
This PR resolves #353 

This PR also:
- updated the doc strings for `tune_band_serial` and `noise_vs_tone` to be consistent with numpy conventions.
- Added convenience function for sample_freq
- Adds testing threading to `profile_band.py`


jesusvasquez333:
This PR resolves #353

PR #373, opened 2020-04-01
Fix noise_vs_bias band requirement, offline mode of analyze_noise_vs_bias, play_tes_bipolar_waveform non-continuous, n_samp for read_stream_data

### LINKED ISSUE #368: `noise_vs_bias` requires band. There is no reason for this.
## Describe the bug
`noise_vs_bias` requires a band number. It should just take data on all the channels and analyze all channels that are on.

eyyoung24:
## Issue
This PR resolves #368 

## Description

- Fixes the band requirement in noise_vs_bias. 
- Allows for offline mode of `analyze_noise_vs_bias`
- Analysis scripts for CMB S4 uMux presentation in `scratch/eyy`
- Adds optional var to play TES waveform continuously or once in `play_tes_bipolar_waveform`
- Pass `n_samp` to `read_stream_data_gcp_save` in `read_stream_data` so the user can limin the number of samples read.

- [x] Yes
- [ ] No

## Which interface changed?
`analyze_slow_iv_from_file` optional var `chs` is changed to `channel` to be more consistent with the rest of the code base.


eyyoung24:
@swh76 I know you're going to dislike the hardcoded offline mode things. I suggest we put it in for now and immediately create an issue to resolve it. I think the thing that others have suggested that we dump_state and then read it in for offline mode should be implemented. Then we can remove all these hardcodes.

PR #378, opened 2020-04-01
Add the `schema` python package to the `pysmurf.client` docker image

### LINKED ISSUE #374: Include python schema package in pysmurf docker
## Describe the problem
Right now, when pysmurf loads a configuration file, it validates the configuration file using the python schema package.  We don't actually import that package, right now we actually just have a copy of the code in pysmurf instead - see here: 
https://github.com/slaclab/pysmurf/blob/master/python/pysmurf/client/base/schema.py
Maintaining their code in ours is silly (and my fault).

Add the schema package to the pysmurf docker. 

## Additional details
If folks are worried about including this nonstandard dependency, there's already an option to not validate the configuration file here:

https://github.com/slaclab/pysmurf/blob/046bd8d41b9db9d133e72a0c0735a9311a1e3083/python/pysmurf/client/base/smurf_config.py#L28

we could make `validate_config=False` if the schema package isn't available for import.  I'll create a separate issue for that and assign it to me.

jesusvasquez333:
## Issue
This PR resolves #374 .

## Description
This PR adds the [schema](https://pypi.org/project/schema) python package to the `pysmurf.client` docker image, and removes the current local copy of the package that live in the pysmurf repository. I pinned the version to the current stable version `0.7.1`.

I updated the code when the package was used to import the newly installed package, instead of using the, now removed, local copy of it.

I also remove the reference of the local copy from the documentation scripts.   

- [ ] Yes
- [X] No


swh76:
Thanks @jesusvasquez333.  Question though ; if we push this into master now, will it break pysmurf on

jesusvasquez333:
As @swh76 pointed out, this change won't work until we release a new docker image, as all the current images don't have the `schema` package installed. So the `import schema` call will always fail until the next release. So I will add a try-catch statement, and fall back to import the local copy.

jesusvasquez333:
@swh76 I added a backward compatible check: if the `import schema` fails, then the local copy will be used instead. I had to add back the local copy of the `schema` package for this to work. I however, leave it out from the documentation.

Issue #393, opened 2020-04-07
Add python library import order check to flake


swh76:
This simple plugin for flake8 can enforce pep8 module import order standards:

```
pip3 install flake8-import-order
flake8 --import-order-style=pep8 --application-import-names=pysmurf
```

the extension is automatically used if installed.  I'm about to open a PR which puts all our imports in pep8 order, after which we'll pass this check.

PR #394, opened 2020-04-07
Reorders imports in compliance with pep8.

swh76:
## Issue
This PR is required to issue resolve https://github.com/slaclab/pysmurf/issues/393.

## Description
Trivial reordering of python module imports throughout pysmurf, both client and core.  After merging this PR, we'll pass this check:

```
flake8 --import-order-style=pep8 --application-import-names=pysmurf
```

if we add the flake8-import-order extension in the travis flake8 checks.

- [ ] Yes
- [x] No


jesusvasquez333:
I will need to carefully check these changes, and test them to make sure nothing breaks. Also, some new orders don't make much sense to me, so I will need to understand why/if there are needed.

swh76:
@jesusvasquez333 oh yea fine to close if you don't want to incorporate.  The only reason why I did it was because it cuts out a bunch of annoying warnings from a linter I was using (pylint) to find some other problems.  I can also just disable the warnings, so whichever you prefer.

jesusvasquez333:
@swh76 not at all. I just need to test the changes before approving this PR. So, I will need some time to tests this.

swh76:
Np @jesusvasquez333 this is like the least urgent PR possible.

eyyoung24:
This may be exactly what Jesus is asking, but... why? Aren't almost all of these imports independent? Is this just code cleanliness? Or is there actually an edge case we're avoiding?

swh76:
Only what I wrote above ; it's just an easy fix to get rid of a bunch of common linting errors.  I presume there is some reason for the pep8 import standards but I don't know it.

Issue #398, opened 2020-04-08
Determine which bays are on from hardware, not pysmurf cfg


swh76:
## Describe the problem
Currently, pysmurf infers which bays have an AMC plugged in by parsing which bands are defined in the pysmurf cfg file ; see

https://github.com/slaclab/pysmurf/blob/2e37c11d3fd1afc1635b17edca93d277a33e8cdc/python/pysmurf/client/util/smurf_util.py#L3993

This is terrible.

## Describe the solution you'd like

I propose a new function `which_bays` which returns which bays are usable (e.g. have an AMC in them and are configurable) by checking

1. Which if any `--disable-bay` options were passed to the rogue startup script (this info is now accessible as described in this JIRA ticket [ESCRYODET-601](https://jira.slac.stanford.edu/browse/ESCRYODET-601)

2. Which bays have accessible PVs.  @jesusvasquez333 suggested checking if `AMCc/FpgaTopLevel/AppTop/AppCore/MicrowaveMuxCore[N]/enable` where N=bay number, 0 or 1 can be read back through epics.  See this JIRA ticket - [ESCRYODET-352](https://jira.slac.stanford.edu/browse/ESCRYODET-352).

We should also add some printout and checks to compare what the user has in their configuration file to which bays are usable.  E.g. at minimum we should report if a user is trying to configure a band that's not usable.

jesusvasquez333:
@swh76 The options `1` and `2` are redundant. You just need to use one of those 2 approaches. I recommend using `1`

swh76:
Thanks @jesusvasquez333 I implemented in PR#403 using approach #1.

PR #403, opened 2020-04-10
Determine which bays are on from hardware, not pysmurf cfg.

swh76:
## Issue
Resolves issue https://github.com/slaclab/pysmurf/issues/398.

## Description

pysmurf has been guessing at which AMC bays are usable by which bands are present in the pysmurf cfg file.  This PR adds a `which_bays` routine to utils which parses the pysmurf server startup arguments to determine which bays were disabled on startup.

- Adds `get_pysmurf_version`, `get_pysmurf_directory`, `get_smurf_startup_script`, and `get_smurf_startup_args` routines to `smurf_command` module for querying SmurfApplication registers.
- Adds a `which_bays` routine to `smurf_utils` module which determines which AMC bays were enabled on pysmurf server startup by parsing the pysmurf server startup options (using the new `get_smurf_startup_script` routine) for the presence of the `--disable-bay0` and `--disable-bay1` options.
- Adds warnings to `smurf_control.initialize` if either the user has bands defined in their pysmurf cfg file that aren't usable (ie belong to an AMC bay that isn't enabled) or if they are missing the definition of a band that is usable.

- [ ] Yes
- [x] No

### DIFF
--- a/python/pysmurf/client/base/base_class.py
+++ b/python/pysmurf/client/base/base_class.py
@@ -79,15 +79,16 @@ class SmurfBase:
             self.log('Offline mode')
 
 
-        # Setting paths for easier commands - Is there a better way to do this
-        # than just hardcoding paths? This needs to be cleaned up somehow
+        # Setting paths for easier commands - Is there a better way to
+        # do this than just hardcoding paths? This needs to be cleaned
+        # up somehow
 
         self.epics_root = epics_root
 
         self.amcc = self.epics_root + ':AMCc:'
 
         self.smurf_application = self.amcc + 'SmurfApplication:'
-        
+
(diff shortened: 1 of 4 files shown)


python/pysmurf/client/base/base_class.py
```diff
@@ -79,12 +79,16 @@ def __init__(self, log=None, epics_root=None, offline=False,
             self.log('Offline mode')
 
 
-        # Setting paths for easier commands - Is there a better way to do this
-        # than just hardcoding paths? This needs to be cleaned up somehow
+        # Setting paths for easier commands - Is there a better way to
```
eyyoung24, line 82: We've been discussing reading the tree back from rogue and populating this. I hate this hard code.

PR #428, opened 2020-05-07
Create new rogue variable `EnabledBays` with a list of enabled bays

### LINKED ISSUE #427: Reliance of startup args in pysmurf functions
Right now the pysmurf function `which_bays` depends on the smurf_startup arguments to see which bays have been disabled. I'm in the process of moving that sort of configuration for my streaming script into the sys_config file instead of doing everything with command line arguments, which will break this function in kind of a weird and unintuitive way. 

Is there any chance we can change this so that the enabled bays are themselves saved in a rogue variable and avoid using the startup_args in future smurf functions?

jlashner:
## Issue
This PR resolves #427.

## Description
This adds a rogue variable `EnabledBays` to the Common root, which is a list of bays that are enabled, and switches `S.which_bays()` to check this variable instead of checking for `--disable-bayX` in the command line args. I've tested at UCSD.

- [ ] Yes
- [X] No

### DIFF
--- a/python/pysmurf/client/command/smurf_command.py
+++ b/python/pysmurf/client/command/smurf_command.py
@@ -4603,7 +4603,6 @@ class SmurfCommandMixin(SmurfBase):
         self._caput(self.channel_mapper + self._payload_size,
                     payload_size, **kwargs)
 
-    _enabled_bays = self.amcc + "EnabledBays"
     def get_enabled_bays(self, **kwargs):
         """
         Gets list of enabled bays.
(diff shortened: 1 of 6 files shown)


jesusvasquez333:
I changed the title of this PR. The title should be descriptive as they go into the release notes.

jesusvasquez333:
I was a little against the idea to add a new rogue variable, as we can already determine the enabled bays in an indirect way (see option 2 in https://github.com/slaclab/pysmurf/issues/398). However, I see that adding it will make things easier, so lets do it.

However, the new variable should no be defined in the root of the tree. I would have put it under `FpgaTopLevel`, but as that lives in the ZIP file, we will have little control on matching a compatible server/client versions. So, I will move it under the `SmurfApplication` device instead.

jlashner:
Ah I see. The problem that I am trying to describe in #427 is that option 1 described in #398 is what is currently implemented, but that doesn't work for our startup scripts, since we don't pass the --disable-bayX to from the command line, but instead run the `server_common.sh` functions and get the modified arguments from stdout. 

Another option would be switching the `which_bays` function to option 2, but I think just adding a rogue variable is clearer and easier, and option2 would require failed caget attempts, which requires a timeout.

jesusvasquez333:
I moved the `EnabledBays` variable under the `SmurfApplication` device, and updated the client command accordingly. I also changed the variable type from a string to a list on ints.

Issue #449, opened 2020-06-15
get/set flux_ramp_freq has different scaling values


eyyoung24:
## Describe the bug
Matt found an inconsistency in our flux_ramp_freq code

https://github.com/slaclab/pysmurf/blob/fabc41fae8bf748fcd2812857b6f5d21dae0898c/python/pysmurf/client/command/smurf_command.py#L3350 

https://github.com/slaclab/pysmurf/blob/fabc41fae8bf748fcd2812857b6f5d21dae0898c/python/pysmurf/client/command/smurf_command.py#L3340

See the hardcoded scaling values. These values should not be hard-coded into pysmurf.

swh76:
RampMaxCnt is in the 307.2MHz clock domain in FW, so I suspect this should just be 3.072e5 (to report in kHz, not MHz).  Instead of hardcoding, could poll `digitizerFrequencyMHz` register via
https://github.com/slaclab/pysmurf/blob/fabc41fae8bf748fcd2812857b6f5d21dae0898c/python/pysmurf/client/command/smurf_command.py#L1803
which is 614.4MHz, and divide by 2.  @jmdewart says the relationship between the RampMaxCnt rate and the digitizer rate is that the default FW rate is 307.2MHz but you get x2 I and x2 Q samples per clock, which is why the digitizer rate is 614.4MHz (2x the default rate).

mhasself:
There are also docstring issues in that vicinity... some units are stated to be in Hz but are clearly kHz.

PR #451, opened 2020-06-17
Fixes wrong hardcoded clock rate in get_flux_ramp_freq, adds docstrings to get_flux_ramp_freq and set_flux_ramp_freq.

swh76:
## Issue
This PR resolves issue https://github.com/slaclab/pysmurf/issues/449.

## Description

Adds docstrings to the smurf_command submodule `get_flux_ramp_freq` and `get_flux_ramp_freq` functions.  Fixes incorrect hardcoded clock rate assumed for `RampMaxCnt` in `get_flux_ramp_freq` (was 3.0725E5, but should be 3.072E5).  Added new `_ramp_max_cnt_clock_hz` attribute to the `SmurfCommandMixin` class that 
`get_flux_ramp_freq` and `set_flux_ramp_freq` both get the clock rate from.  This is still a hardcode, but at least now they'll both be forced to compute the flux ramp reset rate from the same base clock rate.  We could compute on the fly from the digitizer frequency, but then it's not possible to easily check if the flux ramp rate is an integer divisor of the clock rate due to precision error.

Fixed incorrect units in docstring ; these functions take and return the flux ramp reset rate in kHz, not Hz.  Added check to `set_flux_ramp_freq` ; now warns user if requested flux ramp is not an integer divisor of 307.2MHz, and if it's not, reports what the actual programmed flux ramp rate will be (only rates which integer divide 307.2MHz are possible).  

- [ ] Yes
- [x] No

### DIFF
--- a/python/pysmurf/client/command/smurf_command.py
+++ b/python/pysmurf/client/command/smurf_command.py
@@ -192,7 +192,7 @@ class SmurfCommandMixin(SmurfBase):
         Args
         ----
         \**kwargs
-            Arbitrary keyword arguments.  Passed to directly to the
+            Arbitrary keyword arguments.  Passed directly to the
             `_caget` call.
 
         Returns
@@ -215,7 +215,7 @@ class SmurfCommandMixin(SmurfBase):
         Args
         ----
         \**kwargs
-            Arbitrary keyword arguments.  Passed to directly to the
+            Arbitrary keyword arguments.  Passed directly to the
             `_caget` call.
 
         Returns
@@ -235,7 +235,7 @@ class SmurfCommandMixin(SmurfBase):
         Args
         ----
         \**kwargs
-            Arbitrary keyword arguments.  Passed to directly to the
+            Arbitrary keyword arguments.  Passed directly to the
             `_caget` call.
 
         Returns
@@ -255,7 +255,7 @@ class SmurfCommandMixin(SmurfBase):
         Args
         ----
         \**kwargs
-            Arbitrary keyword arguments.  Passed to directly to the
+            Arbitrary keyword arguments.  Passed directly to the
             `_caget` call.
 
         Returns
@@ -294,7 +294,7 @@ class SmurfCommandMixin(SmurfBase):
         Args
         ----
         \**kwargs
-            Arbitrary keyword arguments.  Passed to directly to the
+            Arbitrary keyword arguments.  Passed directly to the
             `_caget` call.
 
         Returns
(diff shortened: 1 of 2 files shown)


swh76:
@mhasself One question - does OCS use the `set_flux_ramp_freq` function?  As far as I can tell, no pysmurf functions do (`setup_flux_ramp`, where it would make the most sense for the function to be used, appears to independently do the same thing - I may open an issue to use `set_flux_ramp_freq` there instead of recomputing `RampMaxCnt`).  @eyyoung24 do you know why we have the `set_flux_ramp_freq` function (I think it's a good idea, just don't want to break anything)?

Re: bounds checking ; we generally try to avoid bounds checking in pysmurf and let rogue take care of it, since it knows those bounds.  However, users currently won't see an error if they exceed register bounds unless they are looking at the rogue server logs (where an error will be reported).  The reasoning has been that if we change any register bounds, pysmurf is kept as agnostic as possible.  But things are a lot more stable than in the old days so I'm not opposed.  What do you think?  If out of bounds errors are being thrown in rogue is that sufficient for OCS?  If not, what's most useful for OCS - a printed warning?  Or raise an exception?

python/pysmurf/client/command/smurf_command.py
```diff
@@ -3390,6 +3390,7 @@ def flux_ramp_off(self, **kwargs):
         self.set_cfg_reg_ena_bit(0, **kwargs)
 
     _ramp_max_cnt = 'RampMaxCnt'
+    _ramp_max_cnt_clock_hz = 307.2e6
```
eyyoung24, line 3393: Is this a thing we should hard-code into pysmurf? Can this ever change?

swh76: @eyyoung24  good - yea, flagrant violation of my own beliefs.  See reason why in PR description ; hardcoding it gets around a rogue issue which otherwise makes it impossible to check if the requested flux ramp rate integer divides 307.2MHz.  Couldn't think of another way around.  I do think this is unlikely to change, but I also do hate hardcoded numbers.  We could drop reporting if the flux ramp set doesn't match the flux ramp requested if we want to compute instead.

swh76: Actually, I think I just found a fix for this.  Will open a JIRA to try and understand why this happens, though.  Hold on merging @eyyoung24.

swh76: Ok, resolving, let me know if doesn't look good to you.

Issue #402, opened 2020-04-10
Finish migrating smurf cfg params loaded by pysmurf to the SmurfConfigPropertiesMixin class


swh76:
Need to finish migrating the smurf configuration file parameters loaded by pysmurf into the SmurfConfigPropertiesMixin defined here : 
https://github.com/slaclab/pysmurf/blob/master/python/pysmurf/client/base/smurf_config_properties.py

swh76:
Should add what the name of each property is in the pysmurf configuration file.

PR #454, opened 2020-06-19
Migrates remaining pysmurf configuration parameters into the SmurfConfigPropertiesMixin class as properties.

swh76:
## Issue
This PR resolves issue https://github.com/slaclab/pysmurf/issues/402.

## Description

This PR is a part of project https://github.com/slaclab/pysmurf/projects/2 ; it mainly closes out issue https://github.com/slaclab/pysmurf/issues/402 (by migrating the remaining pysmurf cfg parameters into python properties in the SmurfConfigPropertiesMixin class) but also does a bunch of other "improvements" related to the pysmurf configuration files and sphinx documentation.  Here's an attempt at a complete list (in no particular order):

- Switches sphinx flag `napoleon_use_ivar=True` in docs/conf.py ; changes the way class attributes are displayed in the sphinx autodocs (instead of adding sections like functions for each one, adds a bulleted list which is easier to parse and makes it easier to read what's a class member function vs a class attribute).
- Adds intersphinx plugin to the sphinx autodocs, allowing links to external python and numpy code used in pysmurf. 
- Adds the `py:func`, `py:exc`, and `py:class` directives to flake8 rst checks (to allow use of intersphinx without triggering flake8).
- Implements the remaining pysmurf configuration parameters that haven't yet been turned into properties as python properties in the SmurfConfigPropertiesMixin class.
- Adds _reg suffix to the register paths for all Rogue parameters in the smurf_command submodule.  This is necessary to avoid the path attributes colliding with the properties, some of which share the same names.
- Removed all self.config queries in all pysmurf functions.  Cfg parameters are now accessed from the properties, specifically from the internal attributes corresponding to each property (e.g. `self._delta_freq` for the `delta_freq` pysmurf configuration file property, when it's used in the `eta_fit` function).
- `hemt_Vd_series_resistor`, `50K_amp_Vd_series_resistor`, and `pA_per_phi0` now float type if user relies on default values (was int).
- Schema validation now allows parameters it doesn't recognize to be present in the pysmurf configuration file.  This is important because this PR eliminates some unused configuration file parameters that will likely be present in users' configuration files for a while until we get everyone synched up by providing an example cfg file.
- Removed unnecessary `set_select_ramp` and `set_ramp_start_mode` calls in `flux_ramp_setup` ; as far as I can tell, it just polls them and then sets them to the values it reads.
- Fixes subplot indexing bug in scratch/shawn/deterministic_latency.py.
- Removes `epics_root` keyword from base class constructor.  Moves `copy_config_to_properties` into SmurfControl class constructor.
- Removed the following unused (as far as I could tell) pysmurf cfg parameters : `chip_to_freq`, all `smurf2mce` parameters, `static_mask`, `mask_channel_offset`, `tune_band:grad_cut`, `tune_band:amp_cut`, `tune_band:n_samples`, `tune_band:freq_max`, `tune_band:freq_min`, `tune_band:eta_scan_amplitude`, and `flux_ramp:select_ramp`.

- [ ] Yes
- [x] No


python/pysmurf/client/base/base_class.py
```diff
@@ -85,7 +85,8 @@ def __init__(self, log=None, epics_root=None, offline=False,
         # do this than just hardcoding paths? This needs to be cleaned
         # up somehow
 
-        self.epics_root = epics_root
+        if epics_root is not None:
```
eyyoung24, line 88: Do we need a default value for this? What do we want to happen if `epics_root` is None?

python/pysmurf/client/debug/smurf_noise.py
```diff
@@ -639,7 +639,7 @@ def noise_vs(self, band, var, var_range, meas_time=30,
                         np.max(self.get_amplitude_scale_array(band)),dtype=int))
                 self.run_serial_gradient_descent(band)
                 self.run_serial_eta_scan(band)
-                self.tracking_setup(band,lms_freq_hz=self.lms_freq_hz[band],
+                self.tracking_setup(band,lms_freq_hz=self._lms_freq_hz[band],
```
eyyoung24, line 642: This isn't introduced here, but since we're reviewing now - Don't know we want `tracking_setup` to use the lms_freq that the user found before? It should probably load the existing `lms_freq`.

python/pysmurf/client/util/smurf_util.py
```diff
@@ -1393,7 +1385,7 @@ def make_mask_lookup(self, mask_file, mask_channel_offset=0,
 
         for gcp_chan, smurf_chan in enumerate(mask):
             b = int(smurf_chan//512)
-            ch = int((smurf_chan-mask_channel_offset)%512)
+            ch = int((smurf_chan)%512)
```
eyyoung24, line 1388: Should we remove the hardcoded 512?

python/pysmurf/client/base/base_class.py
swh76: Added a comment about how default is taken from cfg file by the preceeding `self.copy_config_to_properties` call (since that's super not obvious) and fixed my logic error so that `epics_root` can't accidentally end up set to None (whoops):
https://github.com/slaclab/pysmurf/pull/454/commits/9d9883ae69f6f6ab855a4a3052ac0901dedd61c1
also changed logic for offline slightly ; now just setting to '' in offline mode even if user provides an `epics_root`.  @eyyoung24 lemme know if that breaks something.

python/pysmurf/client/util/smurf_util.py
swh76: Removed 512 hardcodes in smurf_util submodule - https://github.com/slaclab/pysmurf/pull/454/commits/62df701d2201ca7466a8a8e774956ea5839d831c.

swh76:
> Is there documentation about the new config files? I think I get what's going on, but a small description would be helpful.

@eyyoung24 I have this as an to do in https://github.com/slaclab/pysmurf/projects/2 but wasn't planning to have in this PR.  Was going to add a description in the user section and developers section of the sphinx docs (in developers section, was going to walk through how to add a new pysmurf configuration file property).

PR #456, opened 2020-06-22
Add checks and fixes to help solve configuration issues

jesusvasquez333:
## Issue
This PR resolves [ESCRYODET-667](https://jira.slac.stanford.edu/browse/ESCRYODET-667).

## Description

This PR includes some fixes and new features to help solve the configuration issue described in [ESCRYODET-667](https://jira.slac.stanford.edu/browse/ESCRYODET-667):
- Add extra checking while loading defaults: check the final status of the elastic buffer latency. Also, verify the return value and catch any exception from the `LoadConfig` command.
- Increase the rogue timeout from 1s to 5s.
- Increase the `wait_after` period for the set defaults and read all pysmurf client commands. With the previous periods, the commands were returning before the server finished the operations.
- Increase the timeout between deactivate and activate IPMI commands  during the `rebootFPGA` function.

- [ ] Yes
- [X] No

### DIFF
--- a/python/pysmurf/client/command/smurf_command.py
+++ b/python/pysmurf/client/command/smurf_command.py
@@ -410,7 +410,7 @@ class SmurfCommandMixin(SmurfBase):
         """
         Sets the default epics variables
         """
-        self._caput(self.epics_root + ':AMCc:setDefaults', 1, wait_after=20,
+        self._caput(self.epics_root + ':AMCc:setDefaults', 1, wait_after=30,
             **kwargs)
         self.log('Defaults are set.', self.LOG_INFO)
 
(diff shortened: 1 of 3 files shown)


python/pysmurf/client/command/smurf_command.py
```diff
@@ -410,18 +410,18 @@ def set_defaults_pv(self, **kwargs):
...
     def set_read_all(self, **kwargs):
         """
         ReadAll sends a command to read all register to the pyrogue server
-        Registers must upated in order to PVs to update.
-        This call is necesary to read register with pollIntervale=0.
+        Registers must updated in order to PVs to update.
+        This call is necessary to read register with pollIntervale=0.
         """
-        self._caput(self.epics_root + ':AMCc:ReadAll', 1, wait_after=5,
+        self._caput(self.epics_root + ':AMCc:ReadAll', 1, wait_after=20,
```
eyyoung24, line 424: Do we understand why we need 4x as long a wait?

jesusvasquez333: I found this value empirically using the GUI. I execute the ReadAll command and then measure how long does it take to the GUI to start updating values again. The EPICS server keeps responding during this time, but I think it is better to increase this wait to make sure all the registers have been updated by the `ReadAll` command before continuing.

swh76: @jesusvasquez333 @eyyoung24 @slacrherbst  is it possible to have ReadAll set some flag when it completes instead of having an arbitrary 20 sec wait?  I think this would be a more robust solution if possible, since who can say if this time gets less or more in later releases...

PR #458, opened 2020-06-24
Move the `start_server.sh` argument parsing and usage function to `server_common.sh`

### LINKED ISSUE #434: start_server.sh "no exec" mode
## Describe the problem
Currently to run the smurf-streamer, we run a modified version of the start_server.sh script in a subprocess that runs all of the same `server_common` functions, but instead of running another python script when it finishes, it just returns the modified command line args through stdout instead of running another python file. 

It would be great if instead of using our own shell script, we could add a `--no-exec` option to the official one that just prints the command line arguments instead of starting a python script.

jesusvasquez333:
## Issue
This PR resolves #434 

## Description

This PR:
- moves the argument parsing code from the `start_server.sh` script to the `server_common.sh`, to a function called `arg_parser()`. 
- moves the `usage()` function from the `start_server.sh` script to the `server_common.sh`.
- Updates the `start_server.sh` script to call these function from `server_common.sh`.

With these changes, other user can write their own `start_server.sh` function while still using all the common function we keep updated in `server_common.sh`

- [ ] Yes
- [X] No

### DIFF
--- a/docker/server/scripts/server_common.sh
+++ b/docker/server/scripts/server_common.sh
@@ -19,6 +19,110 @@ trap "echo 'An ERROR was found. Check shelf manager & card state! Aborting...'; 
 # Function definitions #
 ########################
 
+# Usage message
+usage()
+{
+    echo "Start the SMuRF server on a specific board."
+    echo ""
+    echo "usage: ${script_name} [-S|--shelfmanager <shelfmanager_name> -N|--slot <slot_number>]"
+    echo "                      [-a|--addr <FPGA_IP>] [-D|--no-check-fw] <pyrogue_server-args>"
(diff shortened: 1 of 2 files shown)


jlashner:
Sorry I'm getting to this late, but it looks good. One thing is that I thought we also talked about moving the main body the run script, i.e. https://github.com/slaclab/pysmurf/blob/e7c79b9753f795f0f649ce440ac45fdff46b1d47/docker/server/scripts/start_server.sh#L20-L47
into its own function in server_common as well, so we don't diverge on what's actually being called. Is this still possible? Thanks.

jesusvasquez333:
@jlashner you are correct. I forgot to implement that change. I will do it in a different PR

jesusvasquez333:
@jlashner I added the missing change here: https://github.com/slaclab/pysmurf/pull/460

PR #460, opened 2020-06-29
Move all the initialization routines to `server_common.sh`

### LINKED ISSUE #434: start_server.sh "no exec" mode

jesusvasquez333:
## Issue
This PR resolves #434 .

## Description

This PR complements PR #458. It moves all the call to initialization routines to a single function called `initialize` which is defined in `server_common.sh`. The `start_server.sh` script now simply calls this function instead of all the other individual functions. This also allows other users to re-use this function. 

- [ ] Yes
- [X] No


docker/server/scripts/server_common.sh
```diff
@@ -741,3 +741,49 @@ detect_amc_board()
...
+    # Firmware version checking
+    checkFW
+
+    # Do a hard boot, if requested
+    hardBoot
+
+    # Auto-detect hardware type
+    ## Detect type of AMCs, and get specific server startup arguments
+    ## for each specific type and add them to the list of arguments
+    local __extar_amcs_args
```
jlashner, line 777: Is this a typo of extra?

jesusvasquez333: Yeap. Thanks for catching it @jlashner . Fortunately (or unfortunately?) all reference to the variable had the same typo so the final result was correct. I fixed it anyways.

jesusvasquez333:
@jlashner with PR should complete solving #434. So, I'm going to delete your branch [issue434](https://github.com/slaclab/pysmurf/tree/issue434) which is not longer necessary

PR #470, opened 2020-07-19
Discarding the master and slave jargon

bopopescu:
For diversity reasons, it would be nice to try to avoid 'master' and 'slave' terminology in this repository which can be associated to slavery. The master-slave terminology could be problematic for people in several countries which has the history of slavery like Romania, USA and many others. Thank you for considering the proposal. Let me know if any changes in the PR are needed, I would be happy to implement them.


swh76:
@bopopescu Hi- thank you for the suggestion & for bringing this to our attention.  We have discussed this and agree with your proposed change.  Unfortunately this code is an API for a larger code that uses the master/slave convention so eliminating the terminology isn't as straightforward as a search & replace.  However, I have discussed it with the owners of that larger code base, and they are now thinking about how to eliminate the terms there as well.  In the meantime, we will work to eliminate the local use of those terms in this repo where possible ; I've started an issue for that here - https://github.com/slaclab/pysmurf/issues/471.

PR #490, opened 2020-08-06
Update smurf_iv.py

SarahMarieBruno:
Adding bias_line_resistance argument to slow_iv_all command.

## Issue
This PR resolves inability to set bias line resistance. 

## Description

Minor change to pass bias_line_resistance argument from slow_iv_all to analyze_slow_iv_all

- [ ] Yes
- [x ] No

### Which interface changed?
Change to arguments in slow_iv_all and analyze_slow_iv_all

Did not have bias_line_resistance argument available

Allows setting the bias line resistance depending on wiring of the testing setup and local physical measurements


swh76:
@SarahMarieBruno I think maybe you have to be a collaborator for the PR checks to run?  Could you try accepting the invitation to be a collaborator on the pysmurf repo and resubmit the PR?  Please also make the title as descriptive as possible ; "Adding `bias_line_resistance` argument to `slow_iv_all` command." would be perfect as both title and description.

PR #516, opened 2020-09-14
Use S.get_number_sub_bands() in fill_band.py

### LINKED ISSUE #515: fill_band.py has hardcoded subbands
## Describe the problem
fill_band.py fills subbands=np.arange(13,115)

Use S.get_number_sub_bands() and fill them appropriately

jmdewart:
## Issue
This PR resolves #515

### DIFF
--- a/scratch/shawn/fill_band.py
+++ b/scratch/shawn/fill_band.py
@@ -7,7 +7,7 @@ import random
 
 band=int(sys.argv[1])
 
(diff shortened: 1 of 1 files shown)


eyyoung24:
I know this is only in Shawn's scratch directory, but we should probably not merge this in until the low power FW is standard.

swh76 (approved):

PR #536, opened 2020-10-26
Fix SIGABRT on DualDataBuffer destruction

ghallsimpsons:
~DualDataBuffer is called without properly terminating the tx thread,
causing the main thread to receive a SIGABRT.
I proposed a reduction in the tx loop timeout so that the destructor
doesn't take 10 seconds to finish.

## Issue
~DualDataBuffer is called without properly terminating the transmitter thread, causing the main thread to receive a SIGABRT.

## Description
This patch causes the DualDataBuffer destructor to stop the transmitter thread before calling the thread destructor. It also reduces the timeout in the transmitter loop to 1s from 10s, to prevent the DualDataBuffer destructor from waiting up to 10s before being able to join the thread.

- [ ] Yes
- [x] No


jesusvasquez333 (approved):

jesusvasquez333:
Thanks @ghallsimpsons for catching this error. I see now that I defined the flag to stop the thread but I forgot to use it during the object destruction.

PR #542, opened 2020-11-04
Resolving bugs in analyze_noise_vs_bias

### LINKED ISSUE #541: Bugs in analyze_noise_vs_bias
## Describe the bug
The analysis script fails if your provide an IV curve to convert into NEP. 

## Additional details

eyyoung24:
## Issue
This PR resolves #541 .

## Description

There were several plotting bugs in `analyze_noise_vs_bias` which would cause it to fail.

- [ ] Yes
- [x] No

### DIFF
--- a/python/pysmurf/client/debug/smurf_noise.py
+++ b/python/pysmurf/client/debug/smurf_noise.py
@@ -747,19 +747,33 @@ class SmurfNoiseMixin(SmurfBase):
             Dictionary with IV information for band.
         """
         self.log(f'Extracting IV data from {iv_data_filename}')
+
+        # Load the data from file
         iv_data = np.load(iv_data_filename, allow_pickle=True).item()
-        iv_band_data = iv_data[band]
-        iv_high_current_mode = iv_data['high_current_mode']
-        for ch in iv_band_data:
-            v_bias = iv_band_data[ch]['v_bias']
-            if iv_high_current_mode and not high_current_mode:
-                iv_band_data[ch]['v_bias'] = v_bias*self._high_low_current_ratio
-            elif not iv_high_current_mode and high_current_mode:
-                iv_band_data[ch]['v_bias'] = v_bias/self._high_low_current_ratio
+
+        iv_band_data = {}
+        band = np.ravel(np.array([band]).astype(int))  # cast as array for loop
+
+        for b in band:
+            # Extract band specfic data
+            iv_band_data[b] = iv_data[b]
+            iv_high_current_mode = iv_data['high_current_mode']
+
+            for ch in iv_band_data[b]:
+                # Loop over channels and match IV and data high/low current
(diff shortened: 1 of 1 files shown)


python/pysmurf/client/debug/smurf_noise.py
```diff
@@ -747,26 +747,42 @@ def get_iv_data(self, iv_data_filename, band, high_current_mode=False):
...
+                        self._high_low_current_ratio
+                elif not iv_high_current_mode and high_current_mode:
+                    iv_band_data[b][ch]['v_bias'] = v_bias / \
+                        self._high_low_current_ratio
+
         return iv_band_data
 
 
-    def get_si_data(self, iv_band_data, ch):
+    def get_si_data(self, iv_band_data, b, ch):
```
swh76, line 776: For consistency across pysmurf, I vote we use band instead of b.  Also, having all the bands' data in `iv_band_data` is confusing since I assume the band in the name meant it's for a band.  Since get_si_data and the other functions don't need to know the band, can the functions that call these functions instead just take the iv data for the specific band they're being called on?

python/pysmurf/client/debug/smurf_noise.py
```diff
@@ -777,17 +793,19 @@ def get_si_data(self, iv_band_data, ch):
         si : float
             The responsivity.
         """
-        return iv_band_data[ch]['v_bias'], iv_band_data[ch]['si']
+        return iv_band_data[b][ch]['v_bias'], iv_band_data[b][ch]['si']
 
 
-    def NEI_to_NEP(self, iv_band_data, ch, v_bias):
+    def NEI_to_NEP(self, iv_band_data, b, ch, v_bias):
```
swh76, line 799: Same comment here.

python/pysmurf/client/debug/smurf_noise.py
```diff
@@ -955,14 +1008,18 @@ def analyze_noise_vs_bias(self, bias, datafile, channel=None, band=None,
             gs = GridSpec(n_row, n_col)
             ax_NEI = fig.add_subplot(gs[:h_NEI, :w_NEI])
             ax_NEIwl = fig.add_subplot(gs[:h_NEIwl, w_NEI:w_NEI+w_NEIwl])
+
+            # Estimate NEP if IV given
             if est_NEP:
-                if ch not in iv_band_data:
+                if ch not in iv_band_data[b]:
```
swh76, line 1014: Same here, name of variable is confusing if it's not the iv data for a single band.

swh76 (changes requested):
Found the iv_band_data variables confusing and passing band to functions that don't need to know the band like the responsivity and NEP functions, but I don't feel that strongly about it if you want to just push it into main as is.

swh76 (approved):

Issue #512, opened 2020-09-03
Add a flag to skip IV analysis when running S.run_iv


jseibert575:
Assigning to myself, adding a flag to skip the built-in analysis when running IV curves and just output the raw data.

eyyoung24:
This is great.

I think you can add `analyze=True` to the declaration of the function. Then just add `if analyze:` before this line
https://github.com/slaclab/pysmurf/blob/184271182de63976de401f7562085c9bbf9b5520/python/pysmurf/client/debug/smurf_iv.py#L161

eyyoung24:
And also make it output a warning message to log if `analyze=False`.

PR #548, opened 2020-11-13
Adding flag to skip IV analysis functions after running S.run_iv()

jseibert575:
## Issue
This PR resolves https://github.com/slaclab/pysmurf/issues/512. 

## Description

Adds a flag to the [S.run_iv() function](https://github.com/slaclab/pysmurf/blob/491f56f8884ba1e18747c2f64205f6c93e430592/python/pysmurf/client/debug/smurf_iv.py#L37) in order to skip the analysis functions after taking data. Also issues a warning to the log when analysis is skipped. 

- [ ] Yes
- [x ] No

### Which interface changed?
No interface changed. 

N/A

N/A

Main concern is that this branch (iv-analysis-flag) is based on `issue508` branch, which is being used at UCSD because of a known bug in the firmware interacting with our carrier. I'm not sure what the proper way is to merge the changes here to main so that it merges in cleanly, but also how to pull this and update the `issue508` branch we're using correctly so that the new update is included but without screwing up the bug fix we're using.

### DIFF
--- a/python/pysmurf/client/debug/smurf_iv.py
+++ b/python/pysmurf/client/debug/smurf_iv.py
@@ -163,7 +163,7 @@ class SmurfIVMixin(SmurfBase):
         R_sh=self._R_sh
 
         if do_analysis:
-            self.log('Analysis flag set to false.')
+            self.log(f'Analyzing IV (do_analysis={do_analysis}).')
 
             self.analyze_iv_from_file(fn_iv_raw_data, make_plot=make_plot,
                 show_plot=show_plot, save_plot=save_plot,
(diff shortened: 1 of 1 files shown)


python/pysmurf/client/debug/smurf_iv.py
```diff
@@ -158,11 +161,15 @@ def run_iv(self, bias_groups=None, wait_time=.1, bias=None,
...
 
         R_sh=self._R_sh
-        self.analyze_iv_from_file(fn_iv_raw_data, make_plot=make_plot,
-            show_plot=show_plot, save_plot=save_plot,
-            plotname_append=plotname_append, R_sh=R_sh, grid_on=grid_on,
-            phase_excursion_min=phase_excursion_min, channel=channels,
-            band=band, bias_line_resistance=bias_line_resistance)
+
+        if do_analysis:
+            self.log('Analysis flag set to false.')
```
swh76, line 166: Should this read "Analysis flag is set to true?"

swh76 (changes requested):

swh76 (approved):

PR #552, opened 2020-11-14
PySmurf should never set TestBiasDacCtrlRegCh to 0x0

### LINKED ISSUE #551: PySmurf should never set TestBiasDacCtrlRegCh to 0x0
## Describe the bug
PySmurf should never set TestBiasDacCtrlRegCh to 0x0, this will cause it to rail to -10V.  Power on default is 0xE, enabled is 0x2.  Let's make PySmurf only every set 0x2

## Additional details

jmdewart:
## Issue
This PR resolves #551 .

## Description

PySmurf should never set TestBiasDacCtrlRegCh to 0x0, this will cause it to rail to -10V. Power on default is 0xE, enabled is 0x2. Let's make PySmurf only ever set 0x2 or 0xE.  The command will default to 0x2 if it's anything else.

- [ ] Yes
- [x] No


eyyoung24 (approved):
0x2 is enable. 0xE is disable

python/pysmurf/client/command/smurf_command.py
```diff
@@ -3873,6 +3873,7 @@ def set_rtm_slow_dac_enable(self, dac, val=2, **kwargs):
 
         # only ever set this to 0x2 or 0xE (enable or disable)
         if (val != 0x2) and (val != 0xE):
+            self.log("RTM dac val must be 0x2 or 0xE. Setting to 0x2.")
```
swh76, line 3876: Confusing, let's make it say something like "RTM dac enable val" since otherwise user might think it means RTM DAC data value.

python/pysmurf/client/command/smurf_command.py
```diff
@@ -3927,7 +3927,7 @@ def set_rtm_slow_dac_enable_array(self, val, **kwargs):
             'len(val) must be 32, the number of DACs in hardware.')
 
         # only ever set this to 0x2 or 0xE
-        if np.any(np.logical_or(val != 0x2 , val != 0xE)):
+        if np.any(np.logical_and(val != 0x2 , val != 0xE)):
             self.log("All values in val must be 0x2 or 0xE. " +
```
swh76, line 3931: Can we also clarify here that it's the enable val?

python/pysmurf/client/command/smurf_command.py
swh76: Looking at this more, it's odd that the pysmurf function is called `set_rtm_slow_dac_enable` for this register since it's not exactly an enable.  John gives the register the name DacCtrlReg (see https://confluence.slac.stanford.edu/display/SMuRF/RTM+firmware?preview=/232089354/233309457/Cryo_RTM_setup_note_v2.pdf) and says that register "configures AD5790 analog output."  I think in the DAC datasheet this register is called the "Software Control Register."  We should change pysmurf to match.  E.g. `set_rtm_slow_dac_control` instead of `set_rtm_slow_dac_enable`.

swh76 (changes requested):

python/pysmurf/client/command/smurf_command.py
eyyoung24: Created new issue https://github.com/slaclab/pysmurf/issues/568

swh76 (approved):

PR #553, opened 2020-11-15
Add external trigger (from timing system) to the downsampler

jesusvasquez333:
## Issue
This PR resolves [ESCRYODET-472](https://jira.slac.stanford.edu/browse/ESCRYODET-472).

## Description

This PR adds a new feature to the  downsampler module which allow it to be trigger from the timing system, using the same method used in the old `smurf2mce` application for BICEP: The word 96 in the [header](https://github.com/slaclab/pysmurf/blob/main/README.SmurfPacket.md) (External real time clock from timing system) triggers a release of a frame everytime it changes (following what was done in `smurf2mce`).

The downsampler have now an additional enum variable `TriggerMode` which can be set to:
- `Internal` [0]: This is the way the downsampler has always been used so far. A new frame is released on every specified number of received frames; the number of frames is selected by the variable `factor`. This is the default mode.
- `Timing (BICEP)` [1]: A new frame is release every time the word 96 in the header (External real time clock from timing system) changes.

- [ ] Yes
- [X] No

### Which interface changed?
The interface is backwards compatible, but these variables were added to the `Downsampler` module: 
- `FrameCnt`: Counts the number of released frames,
- `TriggerMode`: Select the trigger mode (0: `Internal` (default), 1: `Timing (BICEP)`).

and the following commands:
- `clearCnt`: Clear the `FrameCnt` variable.


eyyoung24:
Is there documentation on how this has been tested?

jesusvasquez333:
@eyyoung24 This hasn't been tested yet with a real timing system (as it is still not functional). I tested it by injecting a fake (software generated) counter in the header's word 96 which simulates what we should receive from the timing system.

jesusvasquez333:
We now have a working timing system. I setup a carrier to receive timing from the timing system, but looking at the [header](https://github.com/slaclab/pysmurf/blob/main/README.SmurfPacket.md) info, the **External real time clock from timing system** (word 96) is always `0`. I guess we would need also a sync box connected to the timing system in order to be able to test this.

slacrherbst (approved):

PR #574, opened 2020-12-09
Add `jupyterlab` to the client docker image

jesusvasquez333:
## Issue
This PR resolves [ESCRYODET-747](https://jira.slac.stanford.edu/browse/ESCRYODET-747).

## Description

This PR adds the python package `jupyterlab` (version `2.2.9`) to the pysmurf client docker image.

- [ ] Yes
- [X] No

### DIFF
--- a/docker/client/Dockerfile
+++ b/docker/client/Dockerfile
@@ -18,7 +18,8 @@ RUN pip3 install \
     jupyter==1.0.0 \
     pytest==6.0.1 \
     GitPython==3.1.7 \
-    schema==0.7.1
+    schema==0.7.1 \
+    jupyterlab==2.2.9
 
 # Install pysmurf
 ARG branch


eyyoung24:
Resolve whether we have jupyter and/or jupyterlab installed. And if we need both.

jesusvasquez333:
We already have the `jupyter` package as part of the docker image. Mitch said that `jupyter` will work as well, is that is the case them we don't need to add `jupyterlab` as well.

jesusvasquez333:
I converted this PR to a `draft` in order to avoid merging it by mistake if it is really not necessary.

jesusvasquez333:
@jmdewart verified that `jupyter` notebook (which is already available in the docker image) works. So, we don't need to add `jupyterlab`. 

Closing this PR.

PR #577, opened 2020-12-18
SMuRF header: Use Unix Epoch time for word 6

### LINKED ISSUE #505: Description of header word 6 is misleading
README.SmurfPacket.md describes word 6 as "64 bit unix time" / "64 bit unix time nanoseconds".  I have been told by @joseffrisch that "word 6 is in standard 64 bit unix time in nanoseconds from 1970 from server. This is locked to NTP/PTP in the server so on long time scales should be correct."

But the [implementation](https://github.com/slaclab/pysmurf/blob/184271182de63976de401f7562085c9bbf9b5520/include/smurf/core/common/Helpers.h#L73-L74) uses [steady_clock](https://en.cppreference.com/w/cpp/chrono/steady_clock), which is not connected to wall clock time at all.  Rather, it is guaranteed monotonic, which is of course very useful for other reasons.

From an SO perspective, changing the docs is an acceptable solution... we will add a system timestamp in our streaming processor.

jesusvasquez333:
## Issue
This PR resolves #505 

## Description

Use the UNIX Epoch Time (expressed in nanoseconds) in word 6 of the SMuRF header.

- [X] Yes
- [ ] No

### Which interface changed?
Word 6 in the [SMuRF packet header](https://github.com/slaclab/pysmurf/blob/main/README.SmurfPacket.md).

Word 6 contained the time obtained using [steady_clock](https://en.cppreference.com/w/cpp/chrono/steady_clock), which was a monotonic clock that will never be adjusted (in nanoseconds).

Word 6 now contains the time obtained using [system_clock](https://en.cppreference.com/w/cpp/chrono/system_clock) instead, which is the wall clock time from the system-wide realtime clock, converted to time since Epoch in nanoseconds.

### DIFF
--- a/README.SmurfPacket.md
+++ b/README.SmurfPacket.md
@@ -16,7 +16,7 @@ The header is 128-byte long and contains the following information:
 |                |       3           |         1         | Smurf timing Configuration                    | (clock lock / unlock), (external ref, fiber timing ref)
 |                |       4           |         4         | Number of channels of data                    | 32 bit word giving number channels of following data
 |       1        |       8           |        40         | TES DAC 0-15                                  | TES DAC values. 16X 20 bit in 10X32 bit words, or 5X 65 bit
-|       6        |      48           |         8         | 64 bit unix time                              | 64 bit unix time nanoseconds
+|       6        |      48           |         8         | 64-bit epoch UNIX time                        | 64-bit UNIX epoch time, in nanoseconds
 |       7        |      56           |         4         | Flux ramp increment                           | signed 32 bit integer for increment
 |                |      60           |         4         | Flux ramp offset                              | signed 32 it integer for offset
 |       8        |      64           |         4         | Counter 0, since last 1Hz marker              | 32 bit counter since last 1Hz marker
--- a/include/smurf/core/common/Helpers.h
+++ b/include/smurf/core/common/Helpers.h
@@ -67,11 +67,11 @@ namespace helpers
     };
 
     // Get the current time in nanoseconds.
-    // Use the steady clock which guarantees to be monotonically increasing.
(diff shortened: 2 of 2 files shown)


jesusvasquez333:
I tested this change using the [pysmurf-custom-transmitter-example](https://github.com/slaclab/pysmurf-custom-transmitter-example) application to print out the header information. I got this:
```
=====================================
Packet received
=====================================

-----------------------
 HEADER:
-----------------------
Version                = 1
Crate ID               = 1
Slot number            = 2
Timing configuration   = 0
Number of channels     = 1
Unix time              = 1608317501390514708
Flux ramp increment    = 0
Flux ramp offset       = 0
Counter 0              = 0
Counter 1              = 0
Counter 2              = 390709342
Average reset bits     = 0x0000
Frame counter          = 160235
TES relay settings     = 0
External time clock    = 0
Control field          = 0
Test parameters        = 0
Number of rows         = 0
Number of row reported = 0
Row length             = 0
Data rate              = 0
TES Bias values:
0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,

-----------------------
 DATA (up to the first 20 points):
-----------------------
Data[0] = 7050
-----------------------

=====================================
``` 

Now, the UNIX time (word 6) shows: `1608317501390514708` nanoseconds. Using [this online epoch converter](https://www.epochconverter.com), it says that this time:
> Assuming that this timestamp is in nanoseconds (1 billionth of a second):
> GMT: Friday, December 18, 2020 6:51:41.390 PM
> Your time zone: Friday, December 18, 2020 10:51:41.390 AM GMT-08:00
> Relative: 3 minutes ago

PR #597, opened 2021-01-30
Updating smurf_cmd.py

### LINKED ISSUE #596: `smurf_cmd.py` out of date
## Describe the bug
We have changed a lot of the interfaces since we deployed this at pole. `smurf_cmd` is terribly out of date. We need to update basically everything. 

## Additional details

eyyoung24:
## Issue
Fixes #596
`smurf_cmd.py` has not been updated since the pole deployment. Lots of broken functions. 

## Description

- [ ] Yes
- [x] No

### Which interface changed?


eyyoung24:
This also resolves #606

PR #611, opened 2021-02-25
Adds get_tes_bias_high_current function

jlashner:
## Issue
This PR resolves [introduce here issue number(s) or jira ticket(s). If there are no issues for this PR, please first open an issue for it].

## Description

- [ ] Yes
- [ ] No

### Which interface changed?


jesusvasquez333:
Moved this PR to the `develop` branch

PR #623, opened 2021-04-01
Band delay us

jmdewart:
## Issue
This issue deprecates refPhaseDelay, refPhaseDelayFine and lmsDelay.  We've introduced a new PV bandDelayUs which handles those 3 variables under the hood.  This also improves estimate_phase_delay.

## Description

- [x] Yes
- [ ] No

### Which interface changed?
refPhaseDelay, refPhaseDelayFine, and lmsDelay are now deprecated.  These are handled by Rogue in the bandDelayUs PV.

refPhaseDelay, refPhaseDelayFine and lmsDelay in the configuration file

bandDelayUs in the configuration file

### DIFF
--- a/cfg_files/stanford/experiment_fp31_cc03-02_lbOnlyBay0.cfg
+++ b/cfg_files/stanford/experiment_fp31_cc03-02_lbOnlyBay0.cfg
@@ -7,9 +7,7 @@
 	"band_0" : {
 		"iq_swap_in" : 0,
 		"iq_swap_out" : 0,
-		# Measured for 8234f45 by SWH on 8/1/19
-		"refPhaseDelay" : 6,
-		"refPhaseDelayFine" : 16,
+		"bandDelayUs" : 8.8,
 		"trigRstDly" : 15, # 0xF
 		"feedbackEnable": 1,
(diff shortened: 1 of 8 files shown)


jmdewart:
1617292988_b0_delay

New estimate_phase_delay using the bandDelayUs PV

eyyoung24:
@jmdewart Do we want to raise some warning if they use a deprecated function?

jmdewart:
The biggest change for users is updating their config files to use "bandDelayUs" and remove "refPhaseDelay", "refPhaseDelayFine" and "lmsDelay".  I made those optional so nothing is immediately broken (i.e. doesn't force users to make this change).  Eventually we'd remove those from cfg schema and pysmurf control.  Not sure if you want to force users to update or not.

jesusvasquez333:
You should merge this PR into the `develop` branch instead of `main`. 

Also, please one of the labels `bug, enhancement, interface-change` so that the release notes get organized correctly.

jesusvasquez333:
I also added the `interface-change` label, as the PR says that it will break an interface.

eyyoung24 (approved):

PR #636, opened 2021-04-26
TKID mixin file for TKID specific functions

### LINKED ISSUE #635: Add new TKID mixin file in util directory
## Describe the problem
We need a TKID mixin file in the util directory for all TKID specific functions.

Make a new tkid_util.py file containing the TKID mixin class.

## Additional details
N/A

yukanak:
## Issue
This PR resolves #635 

## Description

Add new file with TKID mixin class.

- [ ] Yes
- [X] No

### Which interface changed?
New file in client/util

None

New file in client/util


eyyoung24 (approved):
This looks great. In the future, we should pull out the block that plots resonator circle and streaming iq noise and move it into either SmurfNoise or SmurfTune. It's a generally useful tool for umux as well as TKIDs.

Issue #654, opened 2021-05-12
Add ability to flip phase polarity before running find_freq


eyyoung24:
## Describe the problem
There was a FW bug (that has since been resolved) that flipped the sign of the phase. The phase slip measurement in `find_freq` depends on getting that phase polarity correct. To solve this, we introduce a paramter that simply multiplies the phase by -1 before taking the derivative.

## Additional details
Add any other details about the feature request here.

PR #655, opened 2021-05-12
Adding config_file documentation and phase flipping optional variable in find_freq

### LINKED ISSUE #649: Documentation on configuration files
## Describe the problem
There is no README on configuration files. Create a default config file that "should just work" and write explanation for it. 

## Additional details
Add any other details about the feature request here.

eyyoung24:
## Issue
This PR resolves #649 and #654 .

## Description

This adds documentation for config_files. 
This adds an optional variable that flips the phase before taking a derivative in `find_freq`.

- [ ] Yes
- [x] No

### Which interface changed?

### DIFF
--- a/python/pysmurf/client/tune/smurf_tune.py
+++ b/python/pysmurf/client/tune/smurf_tune.py
@@ -876,7 +876,6 @@ class SmurfTuneMixin(SmurfBase):
         grad = np.convolve(angle, np.repeat([1,-1], grad_kernel_width),
             mode='same')
 
-
         amp = np.abs(resp)
 
         grad_loc = np.array(grad > grad_cut)
--- a/python/pysmurf/client/tune/smurf_tune.py
+++ b/python/pysmurf/client/tune/smurf_tune.py
@@ -801,7 +801,7 @@ class SmurfTuneMixin(SmurfBase):
             band=None, subband=None, make_subband_plot=False,
             subband_plot_with_slow=False, timestamp=None, pad=50, min_gap=100,
             plot_title=None, grad_kernel_width=8, highlight_phase_slip=True,
-            amp_ylim=None):
+                  amp_ylim=None, flip_phase=False, plot_phase=False):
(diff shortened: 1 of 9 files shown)


python/pysmurf/client/tune/smurf_tune.py
```diff
@@ -801,7 +801,7 @@ def find_peak(self, freq, resp, rolling_med=True, window=5000,
             band=None, subband=None, make_subband_plot=False,
             subband_plot_with_slow=False, timestamp=None, pad=50, min_gap=100,
             plot_title=None, grad_kernel_width=8, highlight_phase_slip=True,
-            amp_ylim=None):
+            amp_ylim=None, flip_phase=False, plot_phase=False):
```
jesusvasquez333, line 804: Missing doc string for the new `plot_phase` argument. @eyyoung24 Can you add it?

python/pysmurf/client/tune/smurf_tune.py
```diff
@@ -3686,6 +3709,8 @@ def find_all_peak(self, freq, resp, subband=None, rolling_med=False,
             Whether to make a plot.
         save_plot : bool, optional, default True
             Whether to save the plot to self.plot_dir.
+        show_plot : bool, optional, defaullt False
```
jesusvasquez333, line 3712: @eyyoung24 I think this doc string is using the wrong argument name. I think it should be `plot_phase` instead of `show_plot`?

jesusvasquez333 (changes requested):

python/pysmurf/client/tune/smurf_tune.py
eyyoung24: Added docstring

python/pysmurf/client/tune/smurf_tune.py
eyyoung24: Added clarification on plot_phase.

jesusvasquez333 (approved):

PR #657, opened 2021-05-13
Adds wrapper functions in smurf_commands.py for setting/getting filter for streaming data. Adds convenience functions in smurf_utils.py for setting/getting filter and its gain

### LINKED ISSUE #656: Add PVs - filter coefficients
## Describe the problem
Currently we have to epics.caput to set filter coefficients, see cells 29 and 113 in https://github.com/slaclab/pysmurf/blob/tkid/scratch/mdewart/intro_to_smurf.ipynb. We need wrapper functions to fix this.

Add wrapper functions in smurf_command.py.

## Additional details
N/A

yukanak:
## Issue
This PR resolves #656 

## Description

Adds wrapper functions in smurf_commands.py for setting/getting filter for streaming data. Adds convenience functions in smurf_utils.py for setting/getting filter and its gain.

- [ ] Yes
- [X] No

### Which interface changed?

### DIFF
--- a/python/pysmurf/client/command/smurf_command.py
+++ b/python/pysmurf/client/command/smurf_command.py
@@ -6100,7 +6100,7 @@ class SmurfCommandMixin(SmurfBase):
         -------
         band : int
             The band.
-        
+
         Returns
         -------
         coef_fixed : float
@@ -6131,7 +6131,7 @@ class SmurfCommandMixin(SmurfBase):
         -------
         band : int
             The band.
-        
+
         Returns
         -------
         val : int
@@ -6162,7 +6162,7 @@ class SmurfCommandMixin(SmurfBase):
         -------
         band : int
             The band.
-        
+
         Returns
         -------
         win_fixed : float array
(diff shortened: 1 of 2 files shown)


python/pysmurf/client/command/smurf_command.py
```diff
@@ -6074,3 +6074,49 @@ def get_stream_data_source_period(self, **kwargs):
         """
         return self._caget(self.stream_data_source + self._stream_data_period,
             **kwargs)
+
+    _boxcar_scale = "boxcarScale"
+
+    def set_boxcar_filter_coef_fixed(self, band, coef_fixed, **kwargs):
```
eyyoung24, line 6080: We should put "args" and "returns" in the docstrings.

python/pysmurf/client/util/smurf_util.py
```diff
@@ -4668,3 +4668,73 @@ def find_probe_tone_gap(self, band, n_tone=2, n_scan=5,
...
+        Programs window for streaming data.
+        The max length for the window is 2048 (1.2 kHz minimum streaming rate).
+
+        Args
+        -------
+        band : int
+            The band.
+        win_len : int
+            Window length. This should be 2.4 MHz divided by the streaming rate.
+        window_fn : function, optional, default Hamming
```
eyyoung24, line 4712: We should add some description of where we typically get these. Feel free to even just reference this webpage (or something like it) https://docs.scipy.org/doc/scipy-0.19.1/reference/signal.html#window-functions

PR #663, opened 2021-06-11
speed up make_freq_mask

### LINKED ISSUE #662: Making a frequency mask is very slow for a full ufm
## Describe the bug
Making a frequency mask (for instance in the stream_data_on function) is very slow when running on a full UFM, taking around a minute or so. This is because the current implementation is querying the subband info and channel center for each individual channel instead of using the centerFrequencyArray, which will return the centers for all channels in a band.

jlashner:
## Issue
This PR should greatly speed up the creation of frequency mask by changing the required number of epics get requests from  being proportional to the number of channels to being proportional to the number of bands. This use to take ~minutes for a full UFM.  Resolves #662 .

This has not been tested yet, so I'll test when we have a cold system available (or someone else on SO can test before then) and then we can switch this from a draft to a full PR when it works.
## Description
The current implementation of this function makes a number of epics requests for each channel, which takes a very long time when there are a lot of channels. The new version uses the centerFrequencyArray registers etc. to get info for all channels in a given band at once, which should speed it up a lot.

- [ ] Yes
- [X] No

### DIFF
--- a/python/pysmurf/client/util/smurf_util.py
+++ b/python/pysmurf/client/util/smurf_util.py
@@ -3303,13 +3303,14 @@ class SmurfUtilMixin(SmurfBase):
             An array with frequencies associated with the mask file.
         """
         freqs = np.zeros(len(mask), dtype=float)
-        channels_per_band = self.get_number_channels()
+        bands = mask // 512
+        chans = mask % 512
 
-        # iterate over mask channels and find their freq
-        for i, mask_ch in enumerate(mask):
-            b = mask_ch // channels_per_band
-            ch = mask_ch % channels_per_band
-            freqs[i] = self.channel_to_freq(b, ch)
+        for b in range(8):
+            m = bands == b
+            if not m.any():
+                continue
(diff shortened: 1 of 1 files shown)


jlashner:
Just tested on smurf-srv20 and confirmed the freq mask generated by this branch is the same, while greatly speeding up its generation.

python/pysmurf/client/util/smurf_util.py
```diff
@@ -2270,8 +2270,8 @@ def channel_to_freq(self, band, channel=None, yml=None):
         ----
         band : int
             The band the channel is in.
-        channel : int or None, optional, default none
-            The channel number.
+        channel : int, None, or array
```
swh76, line 2273: Sorry, just to hold to docstring convention, could you just change to `channel : int, None, or array, optional, default None`?

swh76 (changes requested):

jlashner:
Alright, docstrings are fixed!

swh76 (approved):

PR #665, opened 2021-07-15
Adds data_path_id to SmurfControl Args

### LINKED ISSUE #664: Data_path_id that allows us add a uique id to pysmurf output and plot paths.
## Describe the problem
We've encountered a problem where two instances of pysmurf (controlling different slots) that are instantiated at the same time will have colliding output and plot directories. For instance, if we start two pysmurf instances for slots 2 and 3, at the same time, they could possibly both be writing data to the directory: `/data/smurf_data/20210714/1626302915/outputs`. This means that if both pysmurf instances try to write an IV-file within the same second, they will have the same path name and overwrite each other (this has actually happened to princeton).

To avoid this, we would like to be able to add a unique identifier to the data path, that we can set for each slot when instantiating. That way instead of colliding, we would have two separate output directories:
```
slot 2: /data/smurf_data/20210714/crate1slot2/1626302915/outputs
slot 3: /data/smurf_data/20210714/crate1slot3/1626302915/outputs
```

The issue experienced at pton is documented here: https://github.com/simonsobs/sodetlib/issues/136

jlashner:
## Issue
Resolves #664. 

## Description
Adds the `data_path_id` to the smurf_control args, which allows one to set a unique id in the data path to avoid collisions.

- [ ] Yes
- [X] No

### DIFF
--- a/python/pysmurf/client/base/smurf_control.py
+++ b/python/pysmurf/client/base/smurf_control.py
@@ -186,7 +186,7 @@ class SmurfControl(SmurfCommandMixin,
               paths to avoid possible collisions between multiple smurf
               instances running simultaneously. For instance, if set to
               ``crate1slot2`` the outputs directory will be::
-                  ``<data_dir>/<date>/crate1slot2/<ctime>/outputs``
+              ``<data_dir>/<date>/crate1slot2/<ctime>/outputs``
 
         See Also
         --------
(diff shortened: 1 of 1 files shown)


python/pysmurf/client/base/smurf_control.py
```diff
@@ -180,6 +181,12 @@ def initialize(self, data_dir=None, name=None,
               Whether to send messages to the OCS publisher.
         payload_size : int, optional, default 2048
               The payload size to set on setup.
+        data_path_id : str, optional
```
swh76, line 184: Sorry this is dumb, but could you change the docstring to say `data_path_id : str, optional, default None`?  Just to follow docstring convention.

swh76 (changes requested):
See minor comment.

jlashner:
Thanks shawn! Fixed the docstring

swh76 (dismissed):

swh76 (approved):

Issue #652, opened 2021-05-11
Setup function leaves streaming on


jlashner:
## Describe the bug
During S.setup() pysmurf sets `S.set_stream_enable(1)` to start the stream but never closes it. Because all this data is written to G3 files, this means data will be written indefinitely unless someone turns it off.

aer7:
After poking around the codebases, not enabling streaming by default feels more logical.

Places that are now redundant:

https://github.com/simonsobs/sodetlib/blob/1ff24c68ef43c055a71948c9358b7e40991839d8/scripts/setup_pysmurf.py#L10

https://github.com/slaclab/pysmurf/blob/0fef2dda87e6da292485266b3bf630c9b7ca97dc/scratch/shawn/thermal_test.py#L287

Anywhere where disable_streaming is true: https://github.com/slaclab/pysmurf/search?q=disable_streaming

Places that must now explicitly start streaming:

https://github.com/slaclab/pysmurf/blob/b95a6dcb40629a0e16a16318e819960b0bf01c7b/cfg_files/rflab/smurf_startup_6carrier.cfg#L14

PR #693, opened 2021-10-30
Don't start streaming after setting up pysmurf

aer7:
## Issue

#652 

## Description

Jack brought up the fact that pysmurf setup() turns on streaming with set_stream_enable(1). After poking around, I believe the correct way to start streaming data from pysmurf is with stream_data_on(), not set_stream_enable(1). So it feels like a bug that set_stream_enable(1) is in there. After poking through sodetlib and pysmurf it seems that everywhere disables streaming immediately, hinting that it isn't actually desired to start streaming after setup.

P.S. Why is there set_stream_enable(1) and stream_data_on and not just stream_data_on?

- [ ] Yes
- [x] No

### Which interface changed?
NA

NA

NA

### DIFF
--- a/python/pysmurf/client/base/smurf_control.py
+++ b/python/pysmurf/client/base/smurf_control.py
@@ -634,9 +634,6 @@ class SmurfControl(SmurfCommandMixin,
                                  self._fraction_full_scale,
                                  write_log=write_log)
 
-            # Turn on stream enable for all bands
-            self.set_stream_enable(1, write_log=write_log)
-
             # Set payload size and mask to a single channel
             self.set_payload_size(payload_size)
             self.set_channel_mask([0])


swh76:
@aer7 does this address an outstanding issue?  I'm worried changes like this, if we don't fully understand them, could break things in subtle ways we don't anticipate.

aer7:
As far as I know enabling and disabling streaming works as expected for our test scripts, sodetlib, and smurf-streamer. I believe that it is unintentional to start streaming during the pysmurf setup() function, justified by the fact that most codebases disable it immediately after setup(). See,

https://github.com/simonsobs/sodetlib/blob/1ff24c68ef43c055a71948c9358b7e40991839d8/scripts/setup_pysmurf.py#L10

https://github.com/slaclab/pysmurf/search?p=1&q=disable_streaming

I can do more thorough tests to make sure streaming still works after this change.

jlashner:
I appreciate the response to the issue, but I would also be hesitant to remove this line. Especially since disabling the stream after Setup is different then never enabling the stream at all. There must've been a reason it was put there in the first place... like maybe the server expects some data to be sent through at setup for initialization purposes. Really hard to know without asking Ed or Jesus why this was necessary.

I think if you test it well it would probably be fine to merge, but we have already figured out how to work around the issue so I don't think it's all that important for us.

PR #696, opened 2021-11-12
Support timing synchronization from either the backplane or fiber

aer7:
## Issue
NA

## Description

This change permits use of the SMuRF timing crate for receiving timing from the fiber or backplane for the purpose of synchronizing the flux ramp of multiple SMuRF carriers.

This code assumes three use cases for SMuRF carriers: 1. their timing is internal, 2. on slot 2, timing is received by the timing fiber connected to its RTM, 3. not on slot 2, their timing is received from the backplane of the SMuRF crate. To support these three use cases, the pysmurf cfg file was changed to support three configurations in the "timing_reference" key: "ext_ref", "fiber", and "backplane". Configuration on ext_ref is unchanged for the purpose of internal triggering. Configuration on "fiber" now receives the timing signal from the fiber connected on the RTM, and distributes that signal to the rest of the backplane. Configuration on backplane will receive these triggers and lock its flux ramp to that signal.

To test this code, one SMuRF timing crate and two SMuRF carriers C03-25 and C03-26 were configured at each with one AMC. The software used was pysmurf checked out to the develop branch at the time of this writing. C03-25 was configured on "fiber" and C03-26 was configured on "backplane" timing using experiment_fp31_cc03-02_lbOnlyBay0.cfg as the baseline configuration. After setup(), the flux ramp on both carriers were enabled with flux_ramp_on(). The 15 kHz flux ramps were synchronous to within 5 microseconds on average. No other software configuration was necessary for this test. P.S. I'm not sure why 15 kHz was selected by default.

More thorough documentation is available at https://confluence.slac.stanford.edu/display/SMuRF/Software+Timing+Crate

Testing procedures for the timing crate should be made as soon as possible.

### DIFF
--- a/cfg_files/stanford/experiment_fp31_cc03-02_lbOnlyBay0.cfg
+++ b/cfg_files/stanford/experiment_fp31_cc03-02_lbOnlyBay0.cfg
@@ -260,7 +260,13 @@
 	#   panel reference.  (LmkReg_0x0147 : 0x1A).  Also sets
 	#   flux_ramp_start_mode=0
 	# "backplane" : takes timing from timing master through
-	#   backplane.  Also sets flux_ramp_start_mode=1.
+	#   backplane.  Also sets flux_ramp_start_mode=1. By design
+	#   all SMuRF carrier slots except slot 2 should receive
+	#   this "backplane" configuration.
+	# "fiber": takes timing from the fiber cable connected to 
+	#   the RTM and also distributes it to other carriers through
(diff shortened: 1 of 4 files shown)


aer7:
Note it's common to use "ext_ref" for internal flux ramp resets. That mode will look for a clock signal on the front of the AMC, but won't, and then switch to internal timing.

swh76 (approved):

PR #697, opened 2021-11-22
Add control of the second 50K on the C04

aer7:
## Issue

[ESCRYODET-852](https://jira.slac.stanford.edu/browse/ESCRYODET-852)

## Description

This code change permits use of the second 50K Low Noise Amplifier (LNA) on the C04 cryocard. This code should be backwards compatible with the existing users on the C02 cryocard. No changes should be made to the other three amplifiers, or TES biases. 

The following optional configurations are added.

- smurf_config.py
  - fiftyk2_Vg, default -0.75
  - fiftyk2_gate_dac_num, default 26
  - fiftyk2_gate_bit_to_V, default 3.88e-6
  - fiftyk2_drain_dac_num, default 28
  - fiftyk2_drain_conversion_m, default -0.224968
  - fiftyk2_drain_conversion_b, default 5.59815
  - fiftyk2_Id_offset, default 0
  - fiftyk2_amp_Vd_series_resistor, default 12.0
  - fiftyk2_opamp_gain, default 9.929

The following functions are added.

- smurf_command.py
  - get_50k2_ps_en
  - set_50k2_ps_en
  - get_50k2_bias
  - get_50k2_drain_milliamps
  - get_50k2_gate_enable
  - set_50k2_gate_enable
  - get_50k2_gate_voltage
  - set_50k2_gate_voltage
  - get_50k2_drain_enable
  - set_50k2_drain_enable
  - get_50k2_drain_voltage
  - set_50k2_drain_voltage
- cryo_card.py
  - get_50k2_bias

The get_50k2_drain_voltage and set_50k2_drain voltage is calculated with the following fit. 

## Test log

```
[ 2022-02-16 01:37:34 ]  Done with setup.

In [3]: S.fiftyk2_Vg
Out[3]: -0.75

In [4]: S.fiftyk2_gate_dac_num  
Out[4]: 26

In [5]: S.fiftyk2_drain_dac_num 
Out[5]: 28

In [6]: S.fiftyk2_gate_bit_to_V 
Out[6]: 3.88e-06

In [7]: S.fiftyk2_drain_dac_num 
Out[7]: 28

In [8]: S.fiftyk2_drain_conversion_m  
Out[8]: -0.224968

In [9]: S.fiftyk2_drain_conversion_b  
Out[9]: 5.59815

In [10]: S.fiftyk2_Id_offset 
Out[10]: 0

In [11]: S.fiftyk2_amp_Vd_series_resistor
Out[11]: 12.0

In [12]: S.fiftyk2_opamp_gain
Out[12]: 9.929

In [13]: S.get_50k2_ps_en()  
Out[13]: True

In [14]: S.get_50k2_bias()
Out[14]: 0.8514257812499999

In [15]: S.get_50k2_gate_enable()  
Out[15]: True

In [16]: S.get_50k2_drain_enable() 
Out[16]: True

In [17]: S.get_50k2_drain_voltage()
Out[17]: 5.59815

In [18]: S.get_50k2_gate_voltage() 
Out[18]: 0.0

In [20]: S.C.get_50k2_bias() 
Out[20]: 0.8520703124999999

In [23]: S.set_50k2_gate_voltage(2.02)

In [24]: S.get_50k2_gate_voltage() 
Out[24]: 2.0199978400000003

In [25]: S.set_50k2_drain_voltage(4)  

In [26]: S.get_50k2_drain_voltage()
Out[26]: 4.000003911132813

In [27]: S.get_50k2_drain_milliamps() 
Out[27]: 10.094129255212005

In [28]: S.set_50k2_drain_voltage(5)  

In [29]: S.get_50k2_drain_milliamps() 
Out[29]: 12.777241854668143
```


aer7:
This is the most basic functionality for the two 50Ks on the C04 that we should iron out first. Then I'll add in functionality for the two HEMTs, test compatibility with the C02, and then update the higher level abstractions. Most of these changes should be straightforward.

https://github.com/slaclab/pysmurf/blob/main/python/pysmurf/client/util/smurf_util.py#L2781.

python/pysmurf/client/base/smurf_config.py
```diff
@@ -517,26 +509,56 @@ def represents_int(string):
...
             # for cryostat card revision C02 (PC-248-103-02-C02).  The
             # resistor on that revision of the cryostat card is R54.
             Optional('50K_amp_Vd_series_resistor', default=10.0): And(float, lambda f: f > 0),
-            # 50K amplifier gate voltage, in volts.
-            "LNA_Vg" : Use(float),
+            # Undocumented
+            "50k_Id_offset" : Use(float),
+
+            # 50K2 amplifier gate voltage, in volts.
+            "LNA2_Vg" : Use(float),
```
jlashner, line 545: Are these additionaly cfg fields required for the schema? I think we should make them optional in order to not break pysmurf for existing systems.

python/pysmurf/client/command/smurf_command.py
```diff
@@ -4545,36 +4614,13 @@ def set_hemt_enable(self, disable=False, **kwargs):
...
-
     _hemt_v_reg = 'HemtBiasDacDataRegCh[33]'
 
     def set_hemt_bias(self, val, override=False, **kwargs):
         """
-        Sets the HEMT voltage in units of bits. Need to figure out the
-        conversion into real units.
+        Sets the HEMT voltage in units of bits. This applies
+        to both HEMT1 and HEMT2 because there is only one register
+        to set the bias with, HemtBiasDacDataRegCh[33].
```
jlashner, line 4623: These statements are a bit confusing. Does this mean that the set_hemt_bias function will set the bias for both Hemt1 and Hemt2 to the same value?

jlashner (changes requested):
Most of this looks good to me! the only thing I disagree with is adding in additional required arguments to the config schema (if I'm interpreting the code correctly), and I think we should make the new fields optional with sensible defaults. Whatever changes are made here should not break existing systems or config files. Other than that, this seems fine to me if it's been well tested.

python/pysmurf/client/command/smurf_command.py
aer7: Yeah exactly. There are actually 32 DACs which we can easily interact with using set_rtm_slow_dac_volt, but for some reason there is also address 33 which I don't fully understand, which is why I added that blurb in. I think I should rewrite set_hemt_bias to set set_rtm_slow_dac_volt on the HEMT1 DAC address, and add set_hemt2_bias for the HEMT2 DAC. That should work, unless I'm missing something.

aer7: I'll save those changes for another PR, I should just modify the 50k stuff first

aer7:
Whew, okay, lots of changes. I believe fiftyk2_bit_to_V converts the PIC values to volts, and needs to be measured again, and is deterministic across all C04. Doing that.

PR #703, opened 2022-01-28
Fix issue where AMC LMK doesn't lock to the FPGA clock

aer7:
Problem

The AMC LMK clock receives 122.88 MHz, cleans it up, then spits it
back out to the FPGA. To do this, it locks to one of two clocks,
CLKin0 or CLKin1. 0 is the FPGA's recovered timing clock, and 1 is the
front panel clock. Previously, with fiber timing, the LMK still tries
to lock to the front panel, fails, then free runs, preventing the
tones from locking to the fiber 122.88 MHz.

Solution

Set LMK 0x146 to 0x08 when taking timing from fiber. This also applies
to the backplane case, so set that as well.

Note this could be in the defaults .yml passed to Rogue, then just hit
Rogue with Init. But those defaults .yml files are hard to edit right
now, so edit them dynamically in pysmurf.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,11 @@
-# If these files are modified, tag only those users.
+# @swh76 will be the default owners for everything. Unless
+# a later match takes precedence, they will be requested
+# for review when someone opens a pull request.
+* @aer7 @swh76
+
+# @aer7 and @slacrherbst owns the code related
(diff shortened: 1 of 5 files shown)


aer7:
Our GITHUB_TOKEN disappeared, awesome. Preferably this gets approved anyway.

PR #716, opened 2022-04-15
Add get_timing_fiber_status

aer7:
Add get_timing_fiber_status to smurf_command.py. This is a useful health check for systems connected to the external fiber timing system.


aer7:
Moving to ESCRYODET-472.

PR #719, opened 2022-05-06
Add ability to powercycle slots from pysmurf

aer7:
Because the SMuRF docker containers are networked, we can call SSH into the shelfmanagers directly and run clia commands. This opens the ability to powercycle slots, adjust fan levels, etc. I'm pretty sure sodetlib does something like this.


aer7:
Closing.

aer7:
Closing.

aer7:
Closing.

PR #721, opened 2022-08-26
Adds data_path_id to the tune_dir

jlashner:
Adds `data_path_id` to the tune_dir to make sure different slots have unique paths.

## Description
We just noticed tune data from different slots are all being saved to the same directory. This can be problematic when operating multiple slots simultaneously, as it's possible that different slots have the same timestamp associated with the tune, in which case the tune-files will be overwritten. Additionally, attempting to pull the latest tune_file using `S.load_tune()` will pull the most recent one out of all slots. 

We added `data_path_id` before to make sure `output` and `plot` directories don't conflict across slots, so it makes sense to extend this to the tune directory as well.

## Tests done on this branch
None yet but it seems straight-forward

None

### DIFF
--- a/python/pysmurf/client/base/smurf_control.py
+++ b/python/pysmurf/client/base/smurf_control.py
@@ -242,7 +242,10 @@ class SmurfControl(SmurfCommandMixin,
                 self.output_dir = os.path.join(self.base_dir, self.date,
                                                data_path_id, name, 'outputs')
 
-            self.tune_dir = self._tune_dir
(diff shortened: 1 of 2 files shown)


swh76 (approved):
I confirmed this works off cryostat by running both `setup_notches` and `relock` on fake resonator frequencies generated via `S.freq_resp=S.fake_resonance_dict([4115,4250,4274,4333])`.  Works.

Issue #742, opened 2022-10-31
Should cache PVs instead of caget


swh76:
@jlashner and I have noticed in SAT1 testing at UCSD that over time EPICs calls seem to get slower and slower in ipython sessions.  I noticed that this can be fixed by restarting the ipython session.  Looking at documentation for pyepics, it appears running caget over and over again like we do currently in pysmurf creates more and more PV objects, each of which is has associated callbacks and other overhead.  Looking here : https://cars9.uchicago.edu/software/python/pyepics3/advanced.html they suggest for accessing many channels a number of strategies, including caching PV objects instead of letting `caget` run wild.  Let's try to implement that for `psmurf`.  I will also confirm that all of our `caget` and `caput` calls are going through the `_caget` and `_caput` wrappers in the `smurf_command` module.

PR #744, opened 2022-10-31
Cache PVs

swh76:
## Description

This PR changes the way we caget/caput all EPICs registers to address issue https://github.com/slaclab/pysmurf/issues/742; EPICs PVs are now cached in the `pysmurf.client.SmurfControl` instance instead of every PV request being a new caget/caput call.  This change is being made to solve some issues we ran into running the SO SAT1 systems where PV gets/sets seem to get slower and slower in proportion to the total number of calls (and in the worst cases, stalling out completely).  This change was cribbed from https://cars9.uchicago.edu/software/python/pyepics3/advanced.html, where the `pyepics` developers suggest caching PV objects to avoid cumulative overhead of large numbers of cagets/caputs.

For the majority of registers accessed via `smurf_command` gets and sets, the PV objects are cached in a dictionary called `_pv_cache` in the `SmurfControl` instance:

https://github.com/slaclab/pysmurf/blob/6d57a474d9fbc39fcf189c2a6b05af75bcd7b849/python/pysmurf/client/base/smurf_control.py#L113

the keys in that dictionary are the EPICs paths of the registers, and the values are PV objects instantiated on those paths.  PV objects are only instantiated on the first get or set of a register.  See e.g. 

https://github.com/slaclab/pysmurf/blob/issue742/python/pysmurf/client/command/smurf_command.py#L104-L108

Some special handling was required for the CryoCard class that handles cryostat card PIC gets/sets; here the class variables `readpv` and `writepv` that had held the EPICs path to the cryostat card PIC read/write PV now are instantiated as `pyepics` PV objects, see https://github.com/slaclab/pysmurf/blob/6d57a474d9fbc39fcf189c2a6b05af75bcd7b849/python/pysmurf/client/command/cryo_card.py#L42 and gets/sets are called on those.

No interface changes ; this change should be invisible to the user other than an improvement in performance, particularly for users doing a lot of high cadence register polling.


python/pysmurf/client/command/smurf_command.py
```diff
@@ -32,9 +32,13 @@
 
 class SmurfCommandMixin(SmurfBase):
 
+    # PV cache.  Used to just caget or caput every reg but this is not
+    # efficient.  See
+    # https://cars9.uchicago.edu/software/python/pyepics3/advanced.html.
+    _pv_cache = {}
```
jlashner, line 38: This should go in the SmurfControl `__init__` function since it is not a static variable.

jlashner (changes requested):
One issue is that I think the way the cache is defined right now it will be shared across all pysmurf-instances which I don't think we want. I think the dict initialization needs to go into the SmurfControl `__init__` function.

Other than that, looks good to me as long as you've tested!

python/pysmurf/client/command/smurf_command.py
swh76: Thanks for catching this.  Done!

Issue #724, opened 2022-09-19
No check on drain power supplies before set_amp_defaults in S.setup()


dpdutcher:
https://github.com/slaclab/pysmurf/blob/d7328818f4d2a25d9a8871ef9ceccfa643f55e41/python/pysmurf/client/base/smurf_control.py#L644-L646

According to the `set_amp_defaults` doc string, it assumes the drain power supplies are disabled, but there is no check or command here to ensure that is the case. Running S.setup() with amps that are already powered can have undesirable effects.
Recommend the drain power supplies are disabled prior to this call to `set_amp_defaults`.

dpdutcher:
@swh76 Here's something on the pysmurf front that should be a relatively easy fix.
I don't think it will hamper things in the way SAT-1 testing is proceeding with Jack at the helm, but it did confuse LATR folks quite a bit.

swh76:
Thanks @dpdutcher for reminder.  I'll plan to address this too in upcoming release.

swh76:
A couple of questions on this one.

1) Why is `setup` being run with amps that are already powered?  FYI `setup` must only be run once per board power cycle.  Or is the issue that folks are using `set_amp_defaults` independently of `setup` (which we should make work without causing any problems)?  FYI there is a software reg that records if a system has already been configured (`AMCc:SmurfApplication:SystemConfigured`) but unfortunately because it's a software reg (ie, not defined in fw), it's not persistent if the rogue server is restarted.  So we could add a check for that in `setup` but it wouldn't prevent rerunning setup if the rogue server has been restarted.  Maybe I will do that anyway, with an override flag so users have to force it to rerun `setup` if the system has been configured but the rogue server hasn't been restarted.

2) Looking at `set_amp_defaults` it's not obvious to me why there'd be an issue rerunning this function if the amps are already powered on.  Can you say more about the problems they were seeing?  For the production SO cards, the function runs this logic;

https://github.com/slaclab/pysmurf/blob/54f3e9527b8f83221e05fda005421d62211a6f99/python/pysmurf/client/command/smurf_command.py#L4556

the issue might be the now deprecated `set_rtm_slow_dac_enable(dac_num,0x2)` function calls (low level DAC enable is now handled in `setup` by the call to `set_defaults_pv`, which must not be run twice).

I'll put this on an oscilloscope and see what it's doing, and confirm that removing those `set_rtm_slow_dac_enable` calls doesn't break things (and hopefully fixes this issue).

[...]

PR #754, opened 2022-12-15
Drop deprecated function set_rtm_slow_dac_enable(); make sure setup is only run once

Prakamya01:
## Description

This PR changes blank to address issue #724. We removed instances of the deprecated function set_rtm_slow_dac_enable() and tested the behavior of the hemt1 gate/drain and 50k1 gate/drain amps pre- and post- removal of the function using a multimeter and oscilloscope. In addition, we note that `setup` should only be run once per power board cycle to prevent undesirable effects if the amps are already powered. Now, `setup` checks if the system has already been configured, and you cannot run it again unless you force a rerun of `setup`. 

Removed 3 instances of set_rtm_slow_dac_enable() in `smurf_command.py` as shown below
https://github.com/slaclab/pysmurf/blob/6d57a474d9fbc39fcf189c2a6b05af75bcd7b849/python/pysmurf/client/command/smurf_command.py#L4589

2 are from the function set_amp_defaults(), 1 is from the function set_amp_drain_voltage(self, amp, volt). 

No interface changes ; this change should be invisible to the user.


swh76:
PR in wrong direction

Issue #759, opened 2023-03-17
Add FileWriter:MaxFileSize hook to pysmurf


swh76:
To enable splitting large datasets over many files.  Here's how to set manually;

> caput -S smurf_server_s6:AMCc:SmurfProcessor:FileWriter:MaxFileSize 20000000

this makes 20MB files (the argument seems to be in bytes).  If zero, defaults to just streaming into the same file forever.  Files are named .1, .2, .3, etc..

PR #760, opened 2023-03-17
Add set/get for MaxFileSize to allow splitting long streaming acquisitions over multiple files based on file size.

swh76:
## Description

Adds new `set_max_file_size` and `get_max_file_size` to restore support for splitting streamed data over multiple files of a fixed size.  If set to zero, streams to the same file indefinitely.  Otherwise if nonzero, streams to the set value in number of bytes.  E.g. `set_max_file_size(2000000)` will set the file partition size to 20MB per file.  Files are split by appending an incrementing integer on the end of the usual ctime taggedfile name ; e.g. .1, .2, .3, etc..  Addresses issue #759.

Also removes datafile glob in `read_stream_data` and added logic to find the mask for rolling data files (which only gets written to disk once for MaxFileSize!=0).

### DIFF
--- a/cfg_files/rflab/experiment_rflab_2xLB_backplane.cfg
+++ b/cfg_files/rflab/experiment_rflab_2xLB_backplane.cfg
@@ -1,0 +1,488 @@
+{
+"epics_root" : "test_epics",
+"init": {
+
+	"dspEnable": 1,
+
+	"band_0" : {
+		"iq_swap_in" : 0,
+		"iq_swap_out" : 0,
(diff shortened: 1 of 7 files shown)


jlashner (dismissed):
This looks fine to me! Is there currently support for reading in rotating .dat files?

swh76:
@jlashner right now not doing anything smart with rotating files, just adding the handles to do it.  Your comment reminds me though there's a bug in reading rolling data because of the mask lookup.  Just committed a fix for that.

Issue #792, opened 2024-05-13
Retry on epics failure for cryocard commands


jlashner:
## Describe the problem
Commands to the cryocard sometimes return None, signaling an epics timeout occurred. Unlike epics cagets in the smurf_command module, the cyrocard `do_read` function has no `retry_on_fail`, so if there is an epics failure, due to either hardware connection or server load, it will just return None instead of trying again.

## Describe the solution you'd like
It would be nice if we could have the option to retry_on_fail for cryocard commands.

tristpinsm:
Hi Jack,
I've been looking at `CryoCard.do_read` and it looks like as it is now it will retry, up to 5 times by default, when trying to read from a given address. So I'm wondering if there is somewhere else where this issue may be coming from? Do you have an example of a command that times out?

Also, looking at that code I'm not sure how it behaves if an epics timeout does occur...
```python
#need double write to make sure buffer is updated
self.writepv.put(cmd_make(1, address, 0))
for self.retry in range(0, self.max_retries):
    self.writepv.put(cmd_make(1, address, 0))
    data = self.readpv.get(use_monitor=use_monitor)
    addrrb = cmd_address(data)
    if (addrrb == address):
        return (data)
return (0)

return (self.readpv.get(use_monitor=use_monitor))
```
My understanding is that a timeout would result in `PV.get` returning `None`, which should then raise an exception when the `cmd_address` tries to interpret it as an `int`. (also noting the unreachable return statement at the end)

jlashner:
Ya I think that's what I determined as well looking at this closer... the retry is failing because `cmd_address` cannot handle `None` inputs.

jlashner:
One such failure is documented here: https://github.com/simonsobs/daq-discussions/discussions/91

PR #794, opened 2024-05-17
fix(client.command.cryo_card): Catch EPICS timeout on read and retry.

tristpinsm:
## Description

Addresses issue #792 

## Jira Issue

N/A

## Tests done on this branch

It's a simple enough fix that that testing didn't seem necessary.

N/A

N/A

N/A

### DIFF
--- a/python/pysmurf/client/base/base_class.py
+++ b/python/pysmurf/client/base/base_class.py
@@ -181,7 +181,8 @@ class SmurfBase:
         self.timing_status = self.amctiming + 'TimingFrameRx:'
 
         self.C = CryoCard(self.rtm_spi_cryo_root + 'read',
-                          self.rtm_spi_cryo_root + 'write')
+                          self.rtm_spi_cryo_root + 'write',
+                          log=self.log)
         self.freq_resp = {}
 
         # RTM slow DAC parameters (used, e.g., for TES biasing). The
--- a/python/pysmurf/client/command/cryo_card.py
+++ b/python/pysmurf/client/command/cryo_card.py
@@ -17,6 +17,8 @@
 import time
 import os
 
+from ..base.logger import SmurfLogger
+
 try:
     import epics
 except ModuleNotFoundError:
@@ -30,7 +32,7 @@ def write_csv(filename, header, line):
         f.write(line+'\n')
 
 class CryoCard():
-    def __init__(self, readpv_in, writepv_in):
+    def __init__(self, readpv_in, writepv_in, log=None):
         """
         Interact with the cryocard via the PIC. To interact via the RTM, use SmurfCommandMixin.
         Needs to be compatible with the C02 and C04 cryocards.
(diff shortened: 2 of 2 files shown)


jlashner:
I think we may want to test this... PVs in the smurf_command module use a 10 second timeout, so a retry happens roughly every 10 sec. We may want to implement the same thing here; as things are now I believe retries will happen immediately which I don't think we want.

tristpinsm:
OK, I've added a timeout between restarts and I can try it out on a test system at SLAC this afternoon. Just to confirm that it is retrying as expected when it can't reach the cryo card. However I'm not sure how to determine the appropriate number of retries or time in between. Currently it's set to retry 5 times, at 5s intervals.

tristpinsm:
Alright so I've tested `CryoCard.read_temperature` with no cryo card connected and it's retrying as it should:
```[ 2024-05-17 19:25:14 ]  CryoCard.do_read waiting 5s before retry 1 / 5.
[ 2024-05-17 19:25:19 ]  CryoCard.do_read waiting 5s before retry 2 / 5.
[ 2024-05-17 19:25:24 ]  CryoCard.do_read waiting 5s before retry 3 / 5.
[ 2024-05-17 19:25:29 ]  CryoCard.do_read waiting 5s before retry 4 / 5.
[ 2024-05-17 19:25:34 ]  CryoCard.do_read waiting 5s before retry 5 / 5.
```

However, after those attempts it returns a nonsense value, which is just how the code is designed.

jlashner:
I'm confused how this is working, shouldn't the `continue` bypass the sleep and log? Maybe this is an instance where it does not timeout but the returned address does not match.

I think instead of manually sleeping, it makes more sense to set the timeout in the epics get fn, like here: https://github.com/slaclab/pysmurf/blob/1ff66325bf174f975c241b5cf6922f378dd58b20/python/pysmurf/client/command/smurf_command.py#L208-L211
 5 sec timeout with 5 retries sounds good to me as a default!

tristpinsm:
> I'm confused how this is working, shouldn't the `continue` bypass the sleep and log? Maybe this is an instance where it does not timeout but the returned address does not match.

yeah, that's a mistake...

> I think instead of manually sleeping, it makes more sense to set the timeout in the epics get fn, like here:
> 
> https://github.com/slaclab/pysmurf/blob/1ff66325bf174f975c241b5cf6922f378dd58b20/python/pysmurf/client/command/smurf_command.py#L208-L211
> 
> 
> 5 sec timeout with 5 retries sounds good to me as a default!

ok

jlashner:
This is looking good to me... one more thing though, I think we should raise an exception on failure instead of just returning 0. Can we add that here?

tristpinsm:
> This is looking good to me... one more thing though, I think we should raise an exception on failure instead of just returning 0. Can we add that here?

Yeah, I agree that makes a lot more sense. I wonder if that change may break other code that is relying on the current behaviour though. I guess I can have look and see.

tristpinsm:
@swh76 this one should also be merged for the next release

PR #796, opened 2024-09-09
feat(smurf_util.turn_on_fixed_tones)

tristpinsm:
## Description

Add a function to turn on fixed tones, to facilitate integration into `sodetlib`.

...

## Jira Issue

ESCRYODET-

### DIFF
--- a/python/pysmurf/client/util/smurf_util.py
+++ b/python/pysmurf/client/util/smurf_util.py
@@ -3844,6 +3844,19 @@ class SmurfUtilMixin(SmurfBase):
     # set by array, not by channel
         self.set_amplitude_scale_array(band,new_amplitude_scale_array)
 
+    def turn_on_fixed_tones(self, band, channels, tone_power=None):
+        if tone_power is None:
+            tone_power = self._amplitude_scale[band]
+
+        amp = self.get_amplitude_scale_array(band)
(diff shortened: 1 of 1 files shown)


python/pysmurf/client/util/smurf_util.py
```diff
@@ -3844,6 +3844,19 @@ def turn_off_fixed_tones(self,band):
...
+            tone_power = self._amplitude_scale[band]
+
+        amp = self.get_amplitude_scale_array(band)
+        feedback = self.get_feedback_enable_array(band)
+
+        amp[channels] = tone_power
+        feedback[channels] = 0
+
+        self.set_amplitude_scale_array(band, amp)
+        self.set_feedback_enable_array(band, amp)
```
msilvafe, line 3858: Typo, needs to be `feedback` not `amp` in line 3858

python/pysmurf/client/util/smurf_util.py
```diff
@@ -3844,6 +3844,19 @@ def turn_off_fixed_tones(self,band):
     # set by array, not by channel
         self.set_amplitude_scale_array(band,new_amplitude_scale_array)
 
+    def turn_on_fixed_tones(self, band, channels, tone_power=None):
```
msilvafe, line 3847: What sets the frequency for these channels?

PR #797, opened 2024-09-20
feat(core.server_scripts.Common): Set log level and format.

tristpinsm:
## Description

Add a log handler with timestamps to capture the output of the rogue (and other) logs.

This works, but rogue is extremely verbose at the `INFO` level. It might be useful to set a more relaxed level for just the `cryo-det` modules.

## Jira Issue

ESCRYODET-

### DIFF
--- a/python/pysmurf/client/base/smurf_config.py
+++ b/python/pysmurf/client/base/smurf_config.py
@@ -442,7 +442,7 @@ class SmurfConfig:
 
                 # data_out_mux
                 Optional("data_out_mux",
-                         default=default_data_out_mux_dict[band]) : \
+                         default=default_data_out_mux_dict[band]) :
                 And([Use(int)], list, lambda l: len(l) == 2 and
                     l[0] != l[1] and all(0 <= ll <= 9 for ll in l)),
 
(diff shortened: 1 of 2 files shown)


jlashner (approved):
This looks great! Do you happen to know if warnings are what is already being printed, or if this will introduce new messages into the logs? I am fine with this as long as its not overwhelming, but it may confuse people who are used to looking through smurf-logs to debug.

tristpinsm:
That's a good point. Warning is the default in python, and I haven't seen the level explicitly set in the rogue python code. Looking at the C++ logging it looks like it is set to error by default... So if your impression is that we have been running under that level so far I can change it.

jlashner:
Do you have an example log-file at the warning level that I could look through? I could tell you if that's what we have by default or not.

tristpinsm:
Sure, I can generate that later today

tristpinsm:
I can't run with the full `sodetlib` software stack right now because of the firmware that is deployed on our test system, but just running SMuRF through `setup` to `find_freq` I don't see any warning messages.

tristpinsm:
@swh76 with your stamp of approval I'll merge this and wrap it into the `pysmurf` release

swh76 (approved):

PR #798, opened 2024-09-24
draft: sketched out SO interface CI workflow.

tristpinsm:
## Description

Draft of github CI workflow to enforce compliance with the SO interface document [here](https://www.overleaf.com/project/5e837cac9659910001e5f71e).

This is just a sketch and the code needs to be fleshed out, tested and documented.

## Jira Issue

ESCRYODET-

### DIFF
--- a/tests/ci/check_so_interface.py
+++ b/tests/ci/check_so_interface.py
@@ -1,0 +1,34 @@
+import ast
+
+# functions and their arguments to check
+# specified here: https://www.overleaf.com/project/5e837cac9659910001e5f71e
+# could move this to a file
+interface = {
+    "python/pysmurf/client/tune/smurf_tune.py": {  # file
+        "SmurfTuneMixin": {                        # class
+            "find_freq": [],                       # function and args
+        }
+    },
+}
+
+def _compare_args(ast_args, spec_args):
+    # check the arguments in the code match the specification
+    pass
+
+if __name__ == "__main__":
+    for fname, spec in interface.items():
+        with open(fname, 'r') as fh:
+            tree = ast.parse(fh.read())
+
+        for key in spec:
+            # check this function/class is in the code
+            assert key in tree
+
+            # compare its arguments
+            if isinstance(tree[key], ast.FunctionDef):
+                assert _compare_args(spec[key], tree[key])
+            elif isinstance(tree[key], ast.ClassDef):
+                # could make this recursive maybe
+                for meth in spec[key]:
+                    assert meth in tree[key]
+                    assert _compare_args(spec[key][meth], tree[key][meth])
--- a/.github/workflows/interface_ci.yml
+++ b/.github/workflows/interface_ci.yml
@@ -1,0 +1,15 @@
+
+name: interface-ci
+on: [pull_request]
+
+jobs:
+  check-so-interface:
+    runs-on: ubuntu-latest
+    defaults:
+        run:
+          working-directory: ./tests/ci
+      steps:
+        - name: Check out the repository to the runner
+          uses: actions/checkout@v4
+        - name: Check SO interface spec
+          run: python ./check_so_interface.py


tristpinsm:
I just sketched out how this could work, but this could be a project for Andy.

tests/ci/check_so_interface.py
```diff
@@ -10,42 +10,210 @@
...
+    "run_iv": [
+
+    ], 
+    "analyze_iv": [
+
+    ]
+}
+freeze = {}
+
+with open('tests/ci/frozen_functions.py', 'r') as fh:   #opens file "fname", in read mode 'r', to be used in code as "fh"
```
tristpinsm, line 102: what is going to be stored in this file? Is it a pseudo-python representation of the objects/methods structure that define the interface?

tests/ci/check_so_interface.py
```diff
@@ -10,42 +10,210 @@
...
-                return False
-            if not isinstance(actual.default, ast.Constant) or actual.default.value != expected['default']:
-                return False
-    # Check for *args
-    # Check for **kwargs
+    
     return True
+
 if __name__ == "__main__":   #if this is the main thing being run
     for fname, spec in interface.items():   #loop over file names "fname" as keys and "spec" as values in dictionary called "interface"
```
tristpinsm, line 199: currently `interface` is not defined

tests/ci/check_so_interface.py
```diff
@@ -10,42 +10,210 @@
...
-            assert key in tree
-            # compare its arguments
-            if isinstance(tree[key], ast.FunctionDef):
-                assert _compare_args(spec[key], tree[key])
-            elif isinstance(tree[key], ast.ClassDef):
-                # could make this recursive maybe
-                for meth in spec[key]: #iterating over methods specified in spec[key]
-                    assert meth in tree[key]
-                    assert _compare_args(spec[key][meth], tree[key][meth])
+            notfound = []
```
tristpinsm, line 203: these two lines don't need to be in the context manager

tests/ci/check_so_interface.py
```diff
@@ -10,42 +10,210 @@
...
+            found = {}
+            dump = ast.dump(tree)
+        for node in ast.walk(tree):
```
tristpinsm, line 206: the first layer in the tree is usually going to be made up of `ClassDef` objects (though not always), which in turn contain `FunctionDef`'s. The original sketch had a check for this.

tests/ci/check_so_interface.py
```diff
@@ -104,49 +130,50 @@
     for node in ast.walk(frozen):
         if isinstance(node, ast.FunctionDef) and node.name in spec_args.keys():
             freeze.update({node.name: node})
+print("FROZEN")
```
andy-durham, line 133: There's a lot of these print checks I put along the way to debug, and visualize the flow of my code. I haven't deleted them yet because I wanted to keep them in here to remember good places/dictionaries and lists to check in my code for future debugging. A final version of this code will definitely have these removed.

tests/ci/check_so_interface.py
```diff
@@ -0,0 +1,352 @@
+import ast
+# functions and their arguments to check
+# specified here: https://www.overleaf.com/project/5e837cac9659910001e5f71e
+interface = {
+    "python/pysmurf/client/tune/smurf_tune.py": {  # file
+        "SmurfTuneMixin": {                        # class
+            "methods": [
```
tristpinsm, line 7: what is this `methods` item used for? We probably only want to check for the existence of the methods that are listed in the interface document, and then this list is redundant with the individual method specs.

tests/ci/check_so_interface.py
```diff
@@ -0,0 +1,352 @@
...
+            print(f"Expected: {spec_meths}")
+            return False
+    return True
+
+if __name__ == "__main__":   #if this is the main thing being run
+    for fname, classname in interface.items():   #loop over file names "fname" as keys and "spec" as values in dictionary called "interface"
+        for classname, funcname in classname.items():
+            for funcname in funcname.keys():
+                with open(fname, 'r') as fh:   #opens file "fname", in read mode 'r', to be used in code as "fh"
+                    tree = ast.parse(fh.read()) #parsing contents of file into abstract syntax tree
```
tristpinsm, line 339: might make sense to open the file and parse the tree in the outer loop to avoid redoing it on every iteration needlessly.

tests/ci/check_so_interface.py
```diff
@@ -0,0 +1,352 @@
...
+        if found_meths != spec_meths:
+            print(f"Mismatch in methods for class {classname}")
+            print(f"Found: {found_meths}")
+            print(f"Expected: {spec_meths}")
+            return False
+    return True
+
+if __name__ == "__main__":   #if this is the main thing being run
+    for fname, classname in interface.items():   #loop over file names "fname" as keys and "spec" as values in dictionary called "interface"
+        for classname, funcname in classname.items():
```
tristpinsm, line 336: in general I would suggest not reusing the same variable name like this, just for ease of understanding.

tristpinsm (commented):
This is really starting to take shape! I took a quick look and left a few comments, but I can give it a closer look later on and once we've had a chance to discuss. Have you tried running this on the code?

tests/ci/check_so_interface.py
```diff
@@ -0,0 +1,352 @@
...
+iv_files - generated by run_iv"
+"""
+def compare_args(node, fname, classname, funcname, type):
+    # Check function
+    if type == "isfunc":
+        spec_args = interface[fname][classname][funcname]["args"]
+        spec_defaults = interface[fname][classname][funcname]["defaults"]
+
+        # Extract function arguments from the AST node
+        found_args = [arg.arg for arg in node.args.args]
```
tristpinsm, line 300: This looks good, but you may want to think about some more complicated cases that might occur. See [here](https://docs.python.org/3/library/ast.html#ast.arg) for an example of all the different types of arguments that can occur. We probably don't need to worry about positional-only or kw-only args because that is a pretty unusual pattern, but at least checking for `vararg` and `kwarg` would be good.

tests/ci/check_so_interface.py
```diff
@@ -0,0 +1,352 @@
...
+run_iv #smurf_iv
+analyze_iv #smurf_iv
+
+DATA OUTPUTS TO DISK
+tune files generated when new resonators are found
+channel mapping file format
+.dat noise files - generated by take_stream_data
+iv_files - generated by run_iv"
+"""
+def compare_args(node, fname, classname, funcname, type):
```
tristpinsm, line 293: As a general practice comment, this function would be more portable if it didn't rely on variables that are outside of it. Currently you pass it keys that it will use to address the `interface` dictionary, which is defined outside the function. Instead you could pass it only what it needs to do the comparison -- something like the contents of `interface[fname][classname][funcname]` and the `node`. Also you can probably figure out if you're dealing with a function or a class from the `node` object rather than pass that info as an argument,

tests/ci/check_so_interface.py
```diff
@@ -0,0 +1,352 @@
...
+        spec_meths = interface[fname][classname]["methods"]
+
+        # Extract class methods from the AST node
+        found_meths = []
+        for child in node.body:
+            if isinstance(child, ast.FunctionDef):
+                found_meths.append(child.name)
+
+        # Compare methods to specified methods in interface dict
+        if found_meths != spec_meths:
```
tristpinsm, line 327: In this case we don't care about the order of methods. Consider using the `set` built-in and check that the intersection of the two is complete.

tristpinsm (commented):
Added some more comments to the code.

tests/ci/check_so_interface.py
```diff
@@ -0,0 +1,343 @@
...
+if __name__ == "__main__":   #if this is the main thing being run
+    setspecclass = set()
+    setspecfunc = set()
+    foundclass = set()
+    foundfunc = set()
+    for fname in interface.keys():   #loop over file names "fname" as keys in dictionary called "interface"
+        with open(fname, 'r') as fh:   #opens file "fname", in read mode 'r', to be used in code as "fh"
+            tree = ast.parse(fh.read()) #parsing contents of file into abstract syntax tree
+            dump = ast.dump(tree)
+        for classname in interface[fname].keys():
```
tristpinsm, line 325: The way this loop is structured has some inefficiencies:
- since you have build everything under the assumption that the structure will be like `{classname: {method_name: ...}}`, you could do something simpler than `ast.walk`
- this might make the code easier to understand and reduce the number of loops you do
- ...but we can discuss the details and whether it's worth shifting
```suggestion
        for node in tree:
            if isinstance(node, ast.ClassDef):
                # check if this class is in our spec
                if node.name in interface[fname]:
                    # check the class
                    compare_args(node, interface[fname][node.name]
                    foundclass.append(node.name)
                    # check its methods
                    for child in node.body:
                        if isinstance(child, ast.FunctionDef) and chlid.name in interface[fname][node.name]:
                            compare_args(child, interface[fname][node.name][child.name])
                            foundclass.append(chlid.name)
            elif isinstance(node, ast.FunctionDef):
                # in principle, we could have functions defined in the body (though I don't think we do in this case)
                if node.name in interface[fname]:
                    compare_args(node, interface[fname][node.name]
                    foundfunc.append(node.name)
```

tristpinsm (commented):
Here's a first round of review :)

tests/ci/check_so_interface.py
```diff
@@ -54,45 +69,53 @@
...
         "SmurfCommandMixin": {
             "run_serial_eta_scan": {
                 "args": [
                     'self', 'band', 'sync_group', 'timeout'
                 ],
                 "defaults": [
                     'True', '240'
-                ]
+                ],
+                "kwarg": [
```
tristpinsm, line 81: should there be a duplicate here?

tests/ci/check_so_interface.py
```diff
@@ -276,68 +303,94 @@
...
+    node : an AST tree node
+        Node from the AST tree that pairs with a 
+        given function from the interface dictionary.
+    intdict : A dictionary
+        Dictionary that contains values for 'args',
+        'defaults' and 'kwargs' for a given function
 
+    Returns
+    -------
+    Boolean, or raises an error
```
tristpinsm, line 324: should update the docstring if you are returning an error message. (or if the expected behaviour becomes raising an exception on failure)

tests/ci/check_so_interface.py
```diff
@@ -276,68 +303,94 @@
...
+    found_args = [arg.arg for arg in node.args.args]
+    found_defaults = [ast.unparse(d) if d else None for d in node.args.defaults]
+    has_varargs = node.args.vararg is not None #check for varargs
+    has_kwargs = node.args.kwarg is not None #check for kwargs
+    #Check for varargs
+    if has_varargs:
+        raise NotImplementedError("This code doesn't currently have the functionality to support varargs, " + (node.name) + " has vararg")
+    #If kwargs detected but kwargs not expected
+    if has_kwargs and intdict.get('kwarg') == None:
+        raise Exception("Kwarg expected to be found in "+ (node.name) + " but kwarg not found")
```
tristpinsm, line 339: I think this message is inverted. If we are here then we found a `kwarg` but did not expect to.

tests/ci/check_so_interface.py
```diff
@@ -276,68 +303,94 @@
...
+    #If kwargs not detected but kwargs expected
+    if not has_kwargs and not intdict.get('kwarg') == None:
```
tristpinsm, line 341: because `None` is a unique object in python, it is better practice to check it using, e.g., `x is None` rather than `x == None`. (this is nitpicking)

Here the syntax would be `and intdict.get('kwarg') is not None`

tests/ci/check_so_interface.py
```diff
@@ -276,68 +303,94 @@
...
+        raise Exception("Kwarg not expected to be found in " + (node.name) + " but kwarg detected")
+    # Compare arguments to specified arguments in interface dict
+    if found_args != spec_args:
+        print(f"Mismatch in arguments for function {node.name}")
+        print(f"Found: {found_args}")
+        print(f"Expected: {spec_args}")
+        return "incorrectargs"
```
tristpinsm, line 348: This works for sure, but it could be nicer if there was a more consistent behaviour in how the function behaved on success/failure.

Right now it can fail either by returning `False`, a `str`, or raising an `Exception`. This is not explained in the docstring. From a user's point of view it would be a lot clearer if there was just one behaviour to expect.

Given you would like to pass on some information about the failure, the best solution might be to raise an `Exception`. You already do that for the `kwargs`/`varargs` case, and you could just change these cases to return an `Exception` that contains the message you are currently printing.

See below for suggestions on handling the exceptions.

tests/ci/check_so_interface.py
```diff
@@ -276,68 +303,94 @@
...
-                        assert compare_args(node, intdict, specfunc)
-                        foundfunc.add(specfunc)
+        for node in ast.walk(tree):
+            if isinstance(node, ast.ClassDef):
+                # check if this class is in our spec
+                if node.name in interface[fname]:
+                    # check the class
+                    foundclass.add(node.name)
+                    for child in node.body:
+                        if isinstance(child, ast.FunctionDef) and child.name in interface[fname][node.name]:
```
tristpinsm, line 376: If the only behaviour you expect is an exception to be raised on failure, this can also be simplified a bit using the `try/except`syntax.
```suggestion
                        if isinstance(child, ast.FunctionDef) and child.name in interface[fname][node.name]:
                            try:
                                # this will function will raise on exception on failure
                                assert compare_args(child, interface[fname][node.name][child.name])
                            except Exception as e:
                                # if an exception was raised this block is executed
                                # we raise a new exception with some more context, based on the original one
                                raise Exception("Function check failed in file: " + fname) from e
```
Note that the `raise .. from` syntax will result in the caught exception being printed as well and the new one containing the information about the file name.

tests/ci/check_so_interface.py
```diff
@@ -276,68 +303,94 @@
...
+                    if check == "incorrectdefaults":
+                        raise Exception("Mismatch for defaults in function "+node.name+" occurred in file: "+fname)
+                    foundfunc.add(node.name)
     if setspecclass != foundclass:
-        print(setspecclass-foundclass)
+        raise Exception("The following classes from the specified classes in the interface were not found:\n"+setspecclass-foundclass)
     if setspecfunc != foundfunc:
-        print(setspecfunc-foundfunc)
+        raise Exception("The following functions from the specified functions in the interface were not found:\n"+setspecfunc-foundfunc)
 pass
```
tristpinsm, line 396: this `pass` statement is not needed (probably leftover from the original draft)

tests/ci/check_so_interface.py
```diff
@@ -276,68 +303,94 @@
...
-                        intdict = interface[fname][classname]
-                        assert compare_args(node, intdict, classname)
-                        foundclass.add(classname)
-                for specfunc in interface[fname][classname].keys():
-                    setspecfunc.add(specfunc)
-                    if isinstance(node, ast.FunctionDef) and node.name == specfunc:
-                        intdict = interface[fname][classname][specfunc]
-                        assert compare_args(node, intdict, specfunc)
-                        foundfunc.add(specfunc)
+        for node in ast.walk(tree):
```
tristpinsm, line 369: Isn't `ast.walk` going to iterate through the entire tree hierarchically, including all of the child nodes/functions of the top-level classes? If this is the case, then there will be a lot of redundancy in this loop because you are again looping over the child nodes of every class that you find... This should probably changed to `for node in tree.body` or something along those lines. (But make sure to test this :) )

tristpinsm (commented):
Hi Andy, this is looking great! I think functionally this is all good, but I left some more suggestions and comments that I think could improve the clarity and usability of the code. This is maybe a bit of a pedantic exercise, but hopefully helps you hone those best practices.

Let me know if you have any questions, and definitely if you disagree with any of the suggestions.

tests/ci/check_so_interface.py
andy-durham: This edit works!

tests/ci/check_so_interface.py
```diff
@@ -321,76 +318,67 @@ def compare_args(node, intdict):
...
             elif isinstance(node, ast.FunctionDef):
                 # in principle, we could have functions defined in the body (though I don't think we do in this case)
                 if node.name in interface[fname]:
-                    check = compare_args(node, interface[fname][node.name])
-                    if check == "incorrectargs":
-                        raise Exception("Mismatch for args in function "+node.name+" occurred in file: "+fname)
-                    if check == "incorrectdefaults":
-                        raise Exception("Mismatch for defaults in function "+node.name+" occurred in file: "+fname)
+                    try:
+                        assert compare_args(child, interface[fname][node.name][node.name])
```
tristpinsm, line 377: there should probably be just one indexing of `node.name` here?

tests/.DS_Store
tristpinsm, line 1: remove this before we merge

tests/ci/check_so_interface.py
```diff
@@ -0,0 +1,384 @@
...
+            elif isinstance(node, ast.FunctionDef):
+                # in principle, we could have functions defined in the body (though I don't think we do in this case)
+                if node.name in interface[fname]:
+                    try:
+                        assert compare_args(child, interface[fname][node.name][node.name])
+                    except Exception as e:
+                        raise Exception("Function check failed in file: " + fname) from e
+                    foundfunc.add(node.name)
+    if setspecclass != foundclass:
+        raise Exception("The following classes from the specified classes in the interface were not found:\n"+setspecclass-foundclass)
```
tristpinsm, line 382: doing some quick tests I noticed that this string concatenation fails with
```TypeError: can only concatenate str (not "set") to str```
The below case with functions would also have this issue.

File: tests/ci/check_so_interface.py [hunk 23]
```diff
@@ -0,0 +1,384 @@
...
+                            try:
+                                assert compare_args(child, interface[fname][node.name][child.name])
+                            except Exception as e:
+                                raise Exception("Function check failed in file: " + fname) from e
+                            foundfunc.add(child.name)
+            elif isinstance(node, ast.FunctionDef):
+                # in principle, we could have functions defined in the body (though I don't think we do in this case)
+                if node.name in interface[fname]:
+                    try:
+                        assert compare_args(child, interface[fname][node.name][node.name])
```
tristpinsm, line 377: there should probably be just one indexing of node.name here?

tristpinsm (commented):
Noted a few more small fixes to make, but then we should be good to merge.

tests/ci/check_so_interface.py
andy-durham: Changed to f-strings

tests/.DS_Store
tristpinsm, line 1: this file is still here...

tristpinsm:
@swh76 I can't approve this PR because I was originally the author, but I think it looks good to go. If you'd like to take a look, feel free, but we'll need your approval to merge (squash merge I think in this case).

PR #819, opened 2025-12-12
fix(stream_data_on): Don't crash if cryo card is unavailable.

tristpinsm:
I don't see any other obvious cases where a missing cryo card should not cause a crash, but a more careful audit may be a good idea.


tpsatt:
As written `flux_ramp_ac_dc_relay_status` gets referenced before assignment on the next line. Perhaps before the `try` block:
```
flux_ramp_ac_dc_relay_status = None
```
Then set this and do the logging in the `try` (or something similar).

PR #824, opened 2026-02-10
Set up to publish to PyPI

BrianJKoopman:
## Description

This PR sets up the repo to publish to PyPI as [pysmurf-slac](https://pypi.org/project/pysmurf-slac/). This will allow packages that use `pysmurf` to define it as a dependency and allow `pip` to perform dependency resolution.

Up until commit 9dc85a7e18acfc54814b1a2d36c887054cf6c0a6 setups up the project metadata and new GitHub Actions workflows to build and publish the wheel to PyPI. Since the name 'pysmurf' was already taken on PyPI, we had to switch the 'distribution package' to something different, we went with 'pysmurf-slac'.

Making that change caused `versioneer` some issues -- it stopped picking up the version number when building the wheel. I couldn't figure out how to properly configure this, and `versioneer` isn't very actively maintained, so I started to run into other issues, like errors when `setup.py` was missing. I ended up migrating to my preferred build-backend, `hatchling`. This resulted in a larger PR than I originally intended. Happy to split this up a bit if that's preferred.

In the move to hatchling we drop the vendored `versioneer`, as we no longer need it, `hatch-vcs` now doing the dynamic version determination. It does, however, produce a slightly different version number format. Here's an example:

**versioneer:** 0.11.2+7.ga1e7e91
**hatch-vcs:** 0.11.2.post1.dev7+ga1e7e91

The last thing that came up is in the server and client Dockerfiles, used for testing here. These both were just dropping `pysmurf` in the `PYTHONPATH`. In order to get the dynamically generated `_version.py` file though, we need to actually install the package with `pip`, so 2c908bb49531af23b9c4c2f02084a9bab418d0b3 does that. Tests pass, but let me know if you think this will cause a problem.

### TL;DR
* Updates package metadata -- **explicitly lists supported Python versions at 3.8-3.12, requiring >=3.8** - d4ba14ee999bf6461c71842a69bf5c7ee2e9c6b4
* Switches build-backends from `setuptools` to `hatchling` - 5c91d096eec5dff3bcc7c30fbe8bec7504a8fddf
* Drops the vendored copy of `versioneer` (and other related files/configurations) - 5c91d096eec5dff3bcc7c30fbe8bec7504a8fddf
* Changes how `pysmurf` is installed in the client/server Docker images - 2c908bb49531af23b9c4c2f02084a9bab418d0b3
* Adds workflow job to test wheel builds - 3cdec9d40ff59d24b30994175781c8a030e58a3f 
* Adds workflow job to publish to PyPI when a tagged release is made - 9dc85a7e18acfc54814b1a2d36c887054cf6c0a6

## Jira Issue

I can't create Jira issues.

## Tests done on this branch

I've only run the tests defined in the existing workflows.

No function interface changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [DOE Code](https://www.osti.gov/doecode/biblio/75053)
 
-[![test-or-deploy](https://img.shields.io/github/actions/workflow/status/slaclab/pysmurf/test-or-deploy.yml)](https://github.com/slaclab/pysmurf/actions/workflows/test-or-deploy.yml) [![Documentation Status](https://readthedocs.org/projects/pysmurf/badge/?version=main)](https://pysmurf.readthedocs.io/en/main/?badge=main)
+[![pypi](https://img.shields.io/pypi/v/pysmurf-slac)](https://pypi.org/project/pysmurf-slac/) ![versions](https://img.shields.io/pypi/pyversions/pysmurf-slac) [![test-or-deploy](https://img.shields.io/github/actions/workflow/status/slaclab/pysmurf/test-or-deploy.yml)](https://github.com/slaclab/pysmurf/actions/workflows/test-or-deploy.yml) [![Documentation Status](https://readthedocs.org/projects/pysmurf/badge/?version=main)](https://pysmurf.readthedocs.io/en/main/?badge=main)
 
 The python control software for SMuRF. Includes scripts to do low level
 commands as well as higher level analysis.


tristpinsm (approved):
This looks great! Thanks for doing this :)

I think that installing with pip in the dockers should be fine, because the part that involves C code hasn't changed. We might want to fold that into the build someday, but that would require coordination with some of the downstream packages that depend on it.

I'll make some issues to move to that way of doing it in the SO dockers as well.

PR #872, opened 2026-04-03
fix(_caput): Use var.typeStr metadata for type coercion

### LINKED ISSUE #853: Rogue 6: Handle type checking more reliably
I had to put in a bunch of logic in `caput` to handle different types that trip up rogue 6 register access. It would be a lot nicer to have a more reliable and coherent implementation of this, either in `pysmurf` or `rogue` itself.

ruck314:
## Summary

- Extracts a `coerce_value_for_var(var, val)` helper in
  `python/pysmurf/client/util/tools.py` that uses `var.typeStr` metadata to
  determine expected types, replacing the `var.value()` read-based type
  discovery in `_caput`.
- Adds a `TYPEMAP` dict for the 12 canonical Rogue 6 typeStr values
  (`UInt8/16/32/64`, `Int8/16/32/64`, `Float32`, `Double`, `Bool`, `String`)
  plus a regex-based fallback for arbitrary-width integer (`UInt<N>` /
  `Int<N>`, 1 ≤ N ≤ 64) and `Float64` typeStrs that real cryo-det firmware
  emits (e.g. `UInt1`, `UInt6`, `Float64`). Non-power-of-2 widths round up
  to the smallest containing native numpy dtype, matching what rogue itself
  caches.
- Handles both scalar registers and array registers (`[np]` typeStr
  suffix) through the same code path. Unrecognized typeStrs still raise
  `TypeError` with `var.path` and the bad typeStr in the message.
- Replaces 4 deprecated `np.uint` call sites in `smurf_command.py` with
  `np.uint64`.
- Adds `python/pysmurf/client/test/test_tools.py` with focused unit tests
  for all canonical TYPEMAP entries, the arbitrary-width fallback (scalar
  and `[np]` array forms), the `Float64` alias, and invalid-typeStr
  rejection (e.g. `UInt0`, `UInt65`, `Float16`, `Foo`). No hardware
  required.
- `docker/server/validate.sh`: adds curl `--retry 5 --retry-delay 5
  --retry-all-errors` and an explicit curl-exit-code check on the GitHub
  API calls in `check_if_private_tag_exist` and `check_if_private_asset_exist`,
  so a transient 5xx from `api.github.com` no longer masquerades as a
  missing tag and fails CI.

## Test plan

- [x] `flake8 --count python/` passes with zero errors
- [x] `pytest python/pysmurf/client/test/test_tools.py` passes (no hardware
      required)
- [x] Existing CI test suite passes with no regressions
- [x] `_caput` enum branch (`var.enum is not None`) and command branch
      (`var.isCommand`) are unchanged
- [x] On hardware/emulator: `S._caput` against a known `UInt1`, `UInt6`,
      and `Float64` field (e.g. fields in
      `firmware/python/CryoDet/DspCoreLib/CryoDetCmbHcd/_CryoChannel.py`)
      writes the register without raising

Closes #853

### DIFF
--- a/docker/server/validate.sh
+++ b/docker/server/validate.sh
@@ -52,11 +52,24 @@ check_if_private_tag_exist()
 
     echo "Searching for tag ${tag} in repo $1 with API URL ${api_url} ..."
 
-    # Big blob of JSON with the tags in it.
-    local tags=$(curl \
-      --url $api_url \
+    # Big blob of JSON with the tags in it. Retry on transient errors (e.g.
+    # 5xx, connection resets) so a flaky GitHub API doesn't fail the build.
+    local tags
+    local curl_rc
+    tags=$(curl \
+      --silent \
+      --show-error \
+      --fail \
+      --retry 5 \
+      --retry-delay 5 \
+      --retry-all-errors \
       --header "Authorization: token ${GITHUB_TOKEN}" \
-      --fail)
+      --url $api_url)
+    curl_rc=$?
+
+    if [ "$curl_rc" -ne 0 ]; then
+        error "Failed to query API URL ${api_url} (curl exit ${curl_rc}) for repo $1"
+    fi
 
     # e.g. 0 if no match
     local grep_count=$(echo "$tags" | grep -c "$tag")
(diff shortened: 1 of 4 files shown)


swh76 (dismissed):

ruck314:
@tristpinsm I pushed [9df48af](https://github.com/slaclab/pysmurf/pull/872/commits/9df48af8348935c81166687e043039e1e9b58c80) to address your comment about `The list of typeStr doesn't account for every possible type`

tristpinsm (commented):
Looks great! Nice to see more unit tests in `pysmurf`.

I will test on the B33 system

tristpinsm:
This is still running into issues. The problem is that `typeStr` can be generated dynamically for variables based on the python `BaseVariable` ([here](https://github.com/slaclab/rogue/blob/938c29108a6b382f2e6607b44f3827535e52976d/python/pyrogue/_Variable.py#L401)). This makes it hard to infer the right type in all cases. It may be that adding a case that checks for the `{val.type}{val.shape}` pattern, and falling back on attempting directly casting from the `typeStr`...

I recursively printed all unique value of `typeStr` in the `AMCc.FpgaTopLevel.AppTop.AppCore.SysgenCryo` block:
```
bool
str
int
float
float64(512,)
uint32(512,)
Float64
UInt1
float64(256000,)
UInt32
Int16
UInt6
UInt2
UInt9
UInt16
UInt7
UInt15
UInt8
UInt3
UInt24
```

tristpinsm:
I added some extra cases to fall back on if the variable has a type string that was generated from numpy or python types. I've yet to test it on a live system.

ruck314:
Closing in light of Tristan's refactor proposal. Stage 3 replaces the hand-written smurf_command/base_class register layer (and its docstrings) with runtime introspection and a generated shim, so this change would be discarded. Leaving #853 open to be revisited on the new layer.

PR #873, opened 2026-04-03
fix: Start Xvfb in server docker when GUI flag is used

### LINKED ISSUE #871: Server docker gui broken in rogue 6 releases
## Describe the bug

Server crashing on startup with gui `-g` flag with error:

```
qt.qpa.xcb: could not connect to display :2.0
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, wayland-egl, wayland, wayland-xcomposite-egl, wayland-xcomposite-glx, xcb.

/usr/local/src/pysmurf_utilities/start_server.sh: line 42:    71 Aborted                 (core dumped) ${cmd}
```

May just need to tweak docker-compose.yml.

ruck314:
## Summary
- Install `xvfb` and `x11-utils` in the server Dockerfile
- Auto-start Xvfb in `start_server.sh` when `-g`/`--gui` is passed and no working display is found
- Prevents the Qt/XCB crash that aborts the server on startup with the GUI flag

## Context
The rogue v6 update switched the GUI from `pyrogue.gui` (which supported both Qt and PyDM backends) to exclusively using `pyrogue.pydm.runPyDM()`. PyDM requires a working X display connection, and the server docker doesn't have one configured by default.

The fix is defensive: if X11 forwarding is properly set up (e.g., via docker-compose mounting `/tmp/.X11-unix`), it uses that. Otherwise, Xvfb provides a fallback virtual framebuffer so the server doesn't abort.

Closes #871

## Test plan
- [x] Build server docker image and verify Xvfb packages are installed
- [x] Run server docker with `-g` flag and no DISPLAY — should start Xvfb automatically instead of crashing
- [x] Run server docker with `-g` flag and working X11 forwarding — should use existing display
- [x] Run server docker without `-g` flag — no Xvfb started, no change in behavior

### DIFF
--- a/docker/server/Dockerfile
+++ b/docker/server/Dockerfile
@@ -1,5 +1,11 @@
 FROM ghcr.io/slaclab/smurf-rogue:R5.0.1
 
+# Install Xvfb for virtual display support when running with GUI (-g)
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    xvfb \
+    x11-utils \
+ && rm -rf /var/lib/apt/lists/*
+
 # Copy all firmware related files, which are in the local_files directory
 RUN mkdir -p /tmp/fw/ && chmod -R a+rw /tmp/fw/
 COPY local_files /tmp/fw/
--- a/docker/server/scripts/start_server.sh
+++ b/docker/server/scripts/start_server.sh
@@ -25,6 +25,24 @@ args+=" ${extra_args}"
 
 echo
 
+# If the GUI flag (-g or --gui) is present, ensure a display is available.
+# Start Xvfb if DISPLAY is not set or the X server is not reachable.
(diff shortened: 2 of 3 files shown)


tristpinsm:
I tested and I found that it was crashing trying to bind to display `:0`. I think this is because we run the docker container with `--network host` and that is propagating the host display somehow. Changing to `:99` fixed it.

When I forward X and try to use the GUI it pops up but just lags and is unusable. I don't ever use the GUI however, so that may be a problem with my system. Maybe @swh76 can give it a try?

tristpinsm:
actually it works if I forward with `-Y` instead of `-X`

tristpinsm (approved):

PR #878, opened 2026-04-07
Faster data reader

tristpinsm:
So far I've tested this on a small `.dat` file and confirmed that the result is identical to the current implementation with different values for the `IQ_mode` and `channel` arguments.

I'll do some more testing, but we should consider if we want to support the python-only reader or just remove it and clean up the code.

It would also be useful to double check the build steps now to make sure the cython compilation doesn't cause any issues.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ python/pysmurf/_version.py
 
 # Cython build artifacts
 *.c
+*.cpp
 *.so
 *.html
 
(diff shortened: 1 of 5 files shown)


tristpinsm:
adding @BrianJKoopman to ask if I did anything really bad to the packaging. I've checked that `pip install` works both for wheels and in editable mode. But it's definitely a bit hack-ish.

swh76 (dismissed):

tristpinsm:
OK, I built the client docker again and confirmed that the old and new reader produce exactly the same output with `IQ_mode` in either mode. I also compared reading with and without headers, with and without TES biases.

I'll merge if no objections.

PR #884, opened 2026-04-22
fix(smurf_tune.find_freq): Make output filename more unique

tristpinsm:
Just discovered an issue where it appears that `find_freq` is overwriting and publishing files with the same path, which is causing issues for `suprsync` at site.

`pysmurf` publishes a new file every time you run `find_freq`. There is a separate file for every band, but it is only labelled by the timestamp up to 1s. (it publishes three files -- `resp`, `resonance`, and `freq`) so if you were to run find_freq in sequence over every band and one of those iterations took less than 1s, you could end up with two files with the same path, the second one overwriting the first, but they would both get published.

I'm a bit surprised that `find_freq` can iterate over bands so quickly, so I will try to confirm that this is in fact what is happening.

The solution proposed here is not necessarily robust, but it should make it much less likely for duplicate files to be generated. (you'd have to be running `find_freq` on the same band in a loop).

I also wrapped the `np.savetxt` in a function that will check for an existing file and overwrite only if asked to. Otherwise this will raise an exception. I set it to not overwrite anywhere it is publishing the file.

I searched for any uses of `amp_sweep` in `pysmurf`, `sodetlib` and `sotodlib` in case they are relying on the file naming pattern, but didn't find anything.

### DIFF
--- a/python/pysmurf/client/command/smurf_cmd.py
+++ b/python/pysmurf/client/command/smurf_cmd.py
@@ -19,6 +19,7 @@ import time
 import numpy as np
 
 import pysmurf.client
+from pysmurf.client.util import tools
 
 sys.path.append(os.path.join(os.path.dirname(__file__), '..', '..'))
 
@@ -162,7 +163,7 @@ def set_port(S, slot, port):
         slots = np.append(slots, slot)
         ports = np.append(ports, port)
 
-    np.savetxt(slot_port_file, np.array([slots, ports]).T, fmt='%i %i')
+    tools.save_to_txt(slot_port_file, np.array([slots, ports]).T, overwrite=True, fmt='%i %i')
 
 
 def get_port(S, slot):
--- a/python/pysmurf/client/debug/smurf_noise.py
+++ b/python/pysmurf/client/debug/smurf_noise.py
@@ -256,7 +256,7 @@ class SmurfNoiseMixin(SmurfBase):
                 # outfn = os.path.join(self.plot_dir, save_name)
                 outfn = os.path.join(self.output_dir, save_name)
 
-                np.savetxt(outfn, np.c_[res_freqs, noise_floors[i], f_knees])
+                tools.save_to_txt(outfn, np.c_[res_freqs, noise_floors[i], f_knees])
                 # Publish the data
                 self.pub.register_file(outfn, 'noise_timestream', format='txt')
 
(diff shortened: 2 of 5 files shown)


kmharrington:
Is there a way to edit the filename to add like, 1,2,3 to the end of the name if the file doesn't exist and overwrite is false? That way the two files still get made and no errors are thrown? 

Note on the editing of the file names: 
* I would keep the `<timestamp>_` pattern the same. There is definitely calls of `int( name.strip('_')[0])` in places and I would prefer not to have to find them all.

tristpinsm:
Yeah, it's definitely possible to do a number of things. I added a a band identifier to make the files more unique: `{timestamp}_b{band}_amp_sweep_{}.txt`. The risk is that things downstream rely on the naming pattern in hard to anticipate ways. I didn't find anything when I searched for `amp_sweep`.

I'm happy to append to the filename instead of raising an exception as well, that's a design choice.

kmharrington:
Ahh, I see. Hadn't noticed the band identifier. I am pretty sure we don't have anything using the amplitude sweeps and so this should be fine for us and is pretty likely to stop the overwrites. 

I think it is also true that I'd prefer this code to just "save more files" than to throw an error.

tristpinsm:
> I think it is also true that I'd prefer this code to just "save more files" than to throw an error.

Looking at implementing this, I'm a bit concerned about the unpredictable behavior it could lead to. What if a task is expecting a file to be at a specific path, but actually the file it wrote got given a different name? Do we signal that somehow and track the new filename? This will add some complexity.

Unless we expect to be writing files at a cadence faster than 1s (which the code assumes we do not), then it may be better to simply raise an exception if it does happen and address the issue (in this case the output of `find_freq` for different bands was not being labelled as such).

tristpinsm:
Actually with the LAT having paused observations, I just measured the `find_freq` run time explicitly and it is closer to 30s wall time as I suspected. So I don't think this is happening because of calls to `find_freq` in sequence.

@kmharrington How many files was `suprsync` complaining about? If it's just a few, then it seems more likely to me this might have been an isolated incident of running two instances of `find_freq` simultaneously (I can't think of another explanation) and nothing to do with the upgraded software.

tristpinsm:
I think I finally figured out why this is happening: `estimate_phase_delay` calls `find_freq` twice in quick succession with a sparse frequency sampling, so it is very fast. It looks like this is fast enough in rogue 6 to write two files within 1s.

But this means that there are in fact cases in normal operations where this will happen so raising an exception is not an option (unless we fix the functions to actually use unique filenames).

It's a bit of a kludge to append an extra bit to the filename, but I can't think of anything else that won't risk unintended consequences by changing the structure of the code more significantly...

tristpinsm:
@swh76 what do you think of this change? I tested `find_freq` and that writing to the same path twice will generate a new file name.

PR #890, opened 2026-04-30
fix(set_feedback_limit_khz): correct off-by-2 register scaling (#787)

### LINKED ISSUE #787: Setting feedback limit using `set_feedback_limit_khz` off by 2?
## Describe the bug

I think we may have an off-by-2 in the set_feedback_limit_khz function ;https://github.com/slaclab/pysmurf/blob/19d236caa1867e724762c5d15c32de2889614be6/python/pysmurf/client/util/smurf_util.py#L1983C1-L1983C1

Folks at MIT report despite setting the limit to 1.2 MHz (via `set_feedback_limit_khz(1200)`) seeing tracking limited to +/- 600 kHz.  Naively, I'd expect the maximum possible limit to be 1.2 MHz, the useable subband half width (our polyphase filter bank subband useable width is 2.4 MHz).

Poking at it I noticed that despite the fact that the `feedbackLimit` register is a 16-bit register, setting it via `set_feedback_limit_khz(1200)` and `set_feedback_limit_khz(600)` result in settings of 0x4000 and 0x8000 respectively, or 2^14 and 2^15, respectively.  I'd have expected `set_feedback_limit_khz(1200)` to result in 0xffff, the full 2^16 bit range of the `feedbackLimit` register.  Trying to set it higher leads to no change because of this limit;

https://github.com/slaclab/pysmurf/blob/19d236caa1867e724762c5d15c32de2889614be6/python/pysmurf/client/util/smurf_util.py#L2000

I will need to empirically verify the scaling.

If there is an off-by-2 error it's probably on this line;

https://github.com/slaclab/pysmurf/blob/19d236caa1867e724762c5d15c32de2889614be6/python/pysmurf/client/util/smurf_util.py#L2003

where I'm betting the units of the `feedbackLimit` register should scale inversely to subband_bandwidth/2., not subband_bandwidth, e.g. it should be

```
desired_feedback_limit_dec = np.floor(desired_feedback_limit_mhz/
            ((subband_bandwidth/2.)/2**16.))
```

(although that would lead to 1200 kHz => 2^16 bits which would wrap around to zero.

ruck314:
## Summary

- Fixes #787. The `feedbackLimit` register encodes the limit as a fraction of the per-subband half-width (`digitizer_freq_mhz / n_subbands`), matching the `eta_mag` convention used elsewhere in the same DSP block (`smurf_tune.py:409-413`). The previous formula divided by the full subband bandwidth (`2 * digitizer_freq_mhz / n_subbands`), producing a register value half the intended one — exactly the off-by-2 reported in the issue and consistent with `@swh76`'s observation that the helper caps out at `0x8000` instead of using the full 16-bit range.
- Switch the denominator to `subband_half_width_mhz / 2**16` and clamp the integer to `2**16 - 1` so an input equal to the half-width does not wrap to zero in firmware.
- Fix a copy-paste error in the docstring (`"The band that is to be turned off"`) and document the clamp.

## Behavior change ⚠️

`set_feedback_limit_khz` — and the `feedbackLimitkHz` config key consumed automatically by `SmurfControl` at setup (`smurf_control.py:574`) — will now write register values **~2× larger** than before, yielding tracking that is twice as permissive. Direction is correct (this is the bug fix), but operators who tuned `feedbackLimitkHz` against the buggy formula will see a behavior change at next setup and should review their configs. Every shipped config in `cfg_files/` (typically `225` or `2000` kHz) is affected.

## Verification

Numerical golden values for canonical SMuRF (`digitizer_freq_mhz = 614.4`, `n_subband = 256`, half-width `2.4 MHz`):

| `feedback_limit_khz` | Old register | New register |
| --- | --- | --- |
| 225 | `0x0C00` (3072) | `0x1800` (6144) |
| 600 | `0x2000` (8192) | `0x4000` (16384) |
| 1200 | `0x4000` (16384) | `0x8000` (32768) |
| 2400 (= half-width) | `0x8000` (32768) | `0xFFFF` (65535) |
| 5000 (above cap) | `0x8000` clamp | `0xFFFF` clamp |

The 1200 → `0x4000` data point under the old formula matches what was reported in the issue. The new formula doubles the register value, restoring agreement with the firmware encoding and `@swh76`'s `set_feedback_limit(0, 2**16-1)` workaround.

`flake8 --count python/pysmurf/client/util/smurf_util.py` → 0 errors (matches what the `Flake8 Tests` CI job runs).

## Test plan

- [ ] `flake8` CI job passes
- [ ] Hardware verification by the issue reporter (`@swh76` or the MIT folks): `set_feedback_limit_khz(band, 1200)` followed by `get_feedback_limit(band)` should return ~32768 (was ~16384)
- [ ] Hardware verification: original symptom from #787 (1200 kHz limit yielding only ±600 kHz of tracking) is resolved — should now show ±1200 kHz

## Out-of-scope

- Editing shipped configs in `cfg_files/`. Deployment-specific.
- Adding regression unit tests. Worth doing, but a separate concern; the project has no non-hardware test scaffolding today.
- Touching `set_feedback_limit` in `smurf_command.py`. Unchanged.

Reviewers — feel free to flag any of the above for inclusion in this PR if you'd prefer to address them here rather than in follow-ups.

Closes #787

### DIFF
--- a/python/pysmurf/client/util/smurf_util.py
+++ b/python/pysmurf/client/util/smurf_util.py
@@ -2149,26 +2149,36 @@ class SmurfUtilMixin(SmurfBase):
 
     def set_feedback_limit_khz(self, band, feedback_limit_khz, **kwargs):
         """
-        Sets the feedback limit
+        Sets the feedback limit, which clamps the magnitude of the
+        per-channel tracked-tone deviation before the loop saturates.
+
+        The 16-bit ``feedbackLimit`` register encodes the limit as a
+        fraction of the per-subband half-width (the same convention as
+        ``eta_mag``). Requests larger than the half-width are clamped,
+        and the integer is clamped to ``2**16 - 1`` so an input equal
+        to the half-width does not wrap to zero in firmware.
 
         Args
         ----
         band : int
-            The band that is to be turned off.
+            The 500 MHz band whose feedback limit will be set.
         feedback_limit_khz : float
-            The feedback rate in units of kHz.
+            Maximum permitted tracked-tone deviation, in kHz.
         """
         digitizer_freq_mhz = self.get_digitizer_frequency_mhz(band)
         n_subband = self.get_number_sub_bands(band)
 
-        subband_bandwidth = 2 * digitizer_freq_mhz / n_subband
-        desired_feedback_limit_mhz = feedback_limit_khz/1000.
+        subband_half_width_mhz = digitizer_freq_mhz / n_subband
+        desired_feedback_limit_mhz = feedback_limit_khz / 1000.
 
-        if desired_feedback_limit_mhz > subband_bandwidth/2:
-            desired_feedback_limit_mhz = subband_bandwidth/2
+        if desired_feedback_limit_mhz > subband_half_width_mhz:
+            desired_feedback_limit_mhz = subband_half_width_mhz
 
-        desired_feedback_limit_dec = int(np.floor(desired_feedback_limit_mhz/
-            (subband_bandwidth/2**16.)))
+        desired_feedback_limit_dec = int(np.floor(
+            desired_feedback_limit_mhz / (subband_half_width_mhz / 2**16.)))
+
+        if desired_feedback_limit_dec >= 2**16:
+            desired_feedback_limit_dec = 2**16 - 1
 
         self.set_feedback_limit(band, desired_feedback_limit_dec, **kwargs)
 


ruck314:
Closing as out of date against main. The off-by-2 scaling is still present, but the set_feedback_limit_khz docstring merged since (#891) documents the old scaling, so the fix now needs to cover both and is cleaner redone against current main. Leaving #787 open.

PR #897, opened 2026-04-30
fix(tracking_setup): Default lms_gain to per-band cfg value (#660)

### LINKED ISSUE #660: lms_gain default setting not zero
## Describe the bug
Right now, if you don't supply the lms_gain parameter to tracking_setup(), it defaults to None and then uses the code
```
if lms_gain is None:
    lms_gain = int(9 - np.ceil(np.log2(100/reset_rate_khz)))
```
to set it to four. Since we are supposed to be setting it to zero right now, having the default be something other than zero caused problems during our testing. As long as lms_gain is supposed to be set to zero, the default should probably be zero to avoid mistakes.

ruck314:
## Summary

Closes #660.

When `lms_gain` is not supplied to `tracking_setup()`, the current code computes it from a hardcoded formula (`9 - ceil(log2(100/reset_rate_khz))`) capped at 7 — which evaluates to **4** for the typical 4 kHz flux-ramp rate. Per the issue, the desired default for current operations is **zero**, and silent non-zero defaults caused real testing problems.

## Change

`python/pysmurf/client/tune/smurf_tune.py`, `tracking_setup`:

```diff
 if lms_gain is None:
-    lms_gain = int(9 - np.ceil(np.log2(100/reset_rate_khz)))
-    if lms_gain > 7:
-        lms_gain = 7
+    lms_gain = self._lms_gain[band]
```

This matches the pattern already used **in the same function** for `feedback_gain` / `feedback_start_frac` / `feedback_end_frac` / `lms_freq_hz`, and the `lms_gain` handling in `track_and_check` (`smurf_tune.py:2924`). Sites that want a zero default can now simply set `"lmsGain": 0` in the per-band block of their experiment cfg.

Docstring updated: drop the now-stale `9 - ceil(log2(100/reset_rate_khz))` hint and note the per-band cfg fallback.

Diff: 1 file changed, 3 insertions(+), 6 deletions(-).

## Why config fallback rather than hardcoding `0`

Issue author notes "*as long as `lms_gain` is supposed to be set to zero*, default should be zero" — i.e. the desired value is operationally fixed today but may shift. Routing the default through the per-band cfg key gives sites a single place to pin it (zero today, anything else later) without another pysmurf change, and stays consistent with every other tracking parameter in `tracking_setup`.

## Behavior change ⚠️

For any caller that relies on the old implicit default (most notably `S.tracking_setup(band)` with `lms_gain` left at `None` and `reset_rate_khz=4`), the effective `lms_gain` now comes from `cfg.band_<n>.lmsGain` instead of being computed as `4`. The 37 shipped cfg files in `cfg_files/` currently have `lmsGain` of `5`, `6`, or `7` per band — operators should review their site's `lmsGain` values if they want the SO-recommended zero behaviour. Shipped cfgs are intentionally **left untouched** in this PR (deployment-specific).

## Out of scope

- Editing shipped cfg files in `cfg_files/`. Operators can pin `"lmsGain": 0` per site.
- Changes to `track_and_check` (already uses `self._lms_gain[band]`) or `flux_mod2` (uses `self.get_lms_gain(band)`) — both already config-driven.
- The JIRA dependency `ESCRYODET-825` referenced on the issue. The pysmurf-side fix has no code dependency on it; if that JIRA still needs landing first, please flag in review.

## Test plan

- [x] `flake8 --count python/` → 0 errors (matches the `Flake8 Tests` CI job in `.github/workflows/test-or-deploy.yml`).
- [x] `python3 -c "import ast; ast.parse(...)"` parses clean.
- [ ] CI: flake8, docker definitions, docs build.
- [ ] Hardware verification: with `"lmsGain": 0` set in the relevant `band_#` cfg block, `S.tracking_setup(band)` (no `lms_gain` arg) programs `lmsGain=0` (verifiable via `S.get_lms_gain(band)`).
- [ ] Regression: when `lms_gain` is passed explicitly, the existing "Using LMS gain is now an advanced feature" log message still fires (unchanged).

### DIFF
--- a/python/pysmurf/client/tune/smurf_tune.py
+++ b/python/pysmurf/client/tune/smurf_tune.py
@@ -2544,7 +2544,8 @@ class SmurfTuneMixin(SmurfBase):
             ** Internal register dynamic range adjustment **
             ** Use with caution - you probably want feedback_gain**
             Select which bits to slice from accumulation over
-            a flux ramp period.
+            a flux ramp period.  If None, defaults to the per-band
+            ``lmsGain`` value from the config file (cf. issue #660).
             Tracking feedback parameters are integrated over a flux
             ramp period at 2.4MHz.  The internal register allows for up
             to 9 bits of growth (from full scale).
@@ -2564,8 +2565,6 @@ class SmurfTuneMixin(SmurfBase):
             ceil(log2( (100e3/2.4e6)*(2.4e6/FR_rate) ))
             So 100kHz SQUID throw at 4kHz has bit growth ceil(log2(100e3/4e3)) = 5 bits.
             Try lms_gain = 4.
-
-            This should be approx 9 - ceil(log2(100/reset_rate_khz)) for CMB applications.
 
             Too low of lms_gain will use only a small dynamic range of the streaming
             registers and contribute to incrased noise.
@@ -2588,9 +2587,7 @@ class SmurfTuneMixin(SmurfBase):
         if reset_rate_khz is None:
             reset_rate_khz = self._reset_rate_khz
         if lms_gain is None:
-            lms_gain = int(9 - np.ceil(np.log2(100/reset_rate_khz)))
-            if lms_gain > 7:
-                lms_gain = 7
+            lms_gain = self._lms_gain[band]
         else:
             self.log("Using LMS gain is now an advanced feature.")
             self.log("Unless you are an expert, you probably want feedback_gain.")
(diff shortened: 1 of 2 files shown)


ruck314:
Closing as out of date against main. Main now defaults lms_gain from the per-band config in track_and_check; tracking_setup still derives it from the reset rate, so the remaining fix is a few lines that are simpler to redo against current main. Leaving #660 open.

PR #898, opened 2026-04-30
fix(get_tone_frequency_offset_mhz): Hardcode PFB offsets as offline fallback (#674)

### LINKED ISSUE #674: Loading old debug data from Tracking Setup
## Describe the bug
After attempting to load one of the debug dat files from tracking_setup, I learned that though we are able to read in the f, df, and sync signals, there is no info that tells you how to map those arrays to the correct channel number. From what I can tell, the `decode_data` function relies on the `get_tone_frequency_offset_mhz` function to determine the channel order and correctly map the tracking data to the channel number, but unfortunately, this data is not saved anywhere at the time of tracking and is not available in offline mode, so I don't think it's possible to determine the correct channel order. 

So we are able to see what general tracking statistics look like from old tracking-setup data, but we are unable to go back and look at how well a specific channel was tracking at a certain point in time. I imagine this is something we want to be able to do, so I think we should save the channel order and any other info required to parse debug data in the `take_debug_data` function

## To Reproduce
Try to load a debug .dat file that comes from an older instance of pysmurf.

## Expected behavior
We want to be able to fully recover debug data

ruck314:
## Summary

Closes #674.

In offline mode (`SmurfControl(offline=True)`), `_caget` returns `None`, so `get_tone_frequency_offset_mhz` cannot supply channel ordering. That breaks the chain `decode_data` → `get_processed_channels` → `get_channel_order` → `get_tone_frequency_offset_mhz` (`np.unique(None)` raises) and prevents users from decoding old `tracking_setup` debug `.dat` files — exactly the user-facing symptom in #674.

This PR adds the canonical 512-element PFB subband-center frequency offset array as a class-level numpy constant on `SmurfCommandMixin` and returns a copy of it from `get_tone_frequency_offset_mhz` when `self.offline` is True. The values are taken verbatim from @swh76's [comment](https://github.com/slaclab/pysmurf/issues/674#issuecomment-1597562571) on the issue. The pattern mirrors the existing offline-fallback in `get_number_sub_bands` (smurf_command.py:511) and `get_number_channels` (smurf_command.py:543).

### Why a hardcoded array is acceptable here

@swh76 (issue thread): *\"these frequencies are the polyphase filter bank (PFB) subband center frequency offsets. They are the same for every band, so actually using the above values should be correct, for all bands. We parameterized them by band in pysmurf / rogue only to maintain flexibility in case, for some reason, some future application needed different bands to have different PFBs, but all existing SMuRF implementations use this identical PFB structure for every 500 MHz band.\"*

@jlashner agreed once it was clarified that these are PFB subband centers (firmware-fixed), not resonator centers (which change with usage).

### What this PR does NOT do

- **Save channel metadata in `take_debug_data` output.** Issue #674's primary use case is loading *already-recorded* `.dat` files; a new file format would not help. The hardcoded fallback also makes new files self-describing-by-convention.
- **Touch `decode_data`, `get_channel_order`, or `take_debug_data`.** They are correct as written; only the upstream data dependency was broken.
- **Add a yaml state-dump mechanism.** Mentioned by @swh76 as a possibly-better long-term solution, but the machinery is not in place. Out of scope.

## Diff

`python/pysmurf/client/command/smurf_command.py` — 1 file changed, 75 insertions, 0 deletions:
- New class-level constant `_default_tone_frequency_offset_mhz` (64 rows × 8 = 512 entries) immediately after `_tone_frequency_offset_mhz_reg`.
- New 2-line `if self.offline:` branch at the top of `get_tone_frequency_offset_mhz`. `.copy()` so callers that mutate the array (numpy index assignment, etc.) cannot poison the class constant.
- No other changes.

## Verification

- `flake8 --count python/` → **0 errors** (matches the `Flake8 Tests` job in `.github/workflows/test-or-deploy.yml`).
- Hardcoded array sanity-checked: 512 entries, all unique values; first 4 = `[0.0, 307.2, 153.6, -153.6]`; last 4 = `[152.4, -154.8, 306.0, -1.2]` — matches the issue text exactly.
- Online-mode path unchanged (the `if self.offline` branch is skipped — same `_caget` call as before).

## Test plan

- [x] `flake8 --count python/` passes (CI parity)
- [ ] CI: flake8, docker definitions, docs build
- [ ] Offline-mode smoke test (no hardware required), against an existing `tracking_setup` `.dat` file:
  ```python
  import numpy as np, pysmurf.client
  S = pysmurf.client.SmurfControl(offline=True)
  off = S.get_tone_frequency_offset_mhz(0)
  assert off.shape == (512,) and off.dtype == np.float64
  assert off[0] == 0.0 and off[1] == 307.2
  f, df, strobe = S.decode_data('<path>/<timestamp>.dat')
  assert f.shape[1] == 512  # recast to full channel count, channels addressable by index
  ```
  Expected to reproduce @swh76's monkey-patch result from the issue thread without monkey-patching.
- [ ] Hardware sanity check (any reviewer with a crate): on an online `SmurfControl` instance, `S.get_tone_frequency_offset_mhz(band)` should be element-wise equal to `pysmurf.client.SmurfControl(offline=True).get_tone_frequency_offset_mhz(band)` for every band on standard SO-SMuRF firmware.


ruck314:
Closing in light of Tristan's refactor proposal. It removes hardcoded channel-geometry fallbacks in favour of geometry reported by the platform, so this offline fallback would be deleted. Leaving #674 open.

PR #900, opened 2026-04-30
fix(setup): Disable streaming before setup() returns (#652)

### LINKED ISSUE #652: Setup function leaves streaming on

ruck314:
## Summary

Closes #652.

`S.setup()` calls `S.set_stream_enable(1)` but never disables it before returning. Because the downstream G3 file writers consume the stream, data is written indefinitely until the user explicitly calls `set_stream_enable(0)` (or `stream_data_off()`). External configs and shell scripts have papered over this with a `disable_streaming=true` cfg flag that runs `S.set_stream_enable(0)` after `setup()` from outside pysmurf.

This PR keeps the existing `set_stream_enable(1)` (so any internal setup step that may implicitly rely on streaming being on still sees it on) and adds a matching `set_stream_enable(0)` immediately before `setup()` returns success. Streaming-during-setup is now symmetric: enabled across the configuration window, disabled on exit. The docstring is updated to describe the new behavior and to direct callers to `stream_data_on()` / `stream_data_off()` for data acquisition.

This is intentionally more conservative than [#693](https://github.com/slaclab/pysmurf/pull/693), which removed the enable line entirely and was closed over reviewer concern that some internal step might rely on streaming being on during setup.

## Out of scope

- Not removing the inline `set_stream_enable(1)` (the PR #693 approach).
- Not adding a new `setup()` parameter.
- Not modifying any cfg file or the external `disable_streaming` workaround machinery — those become redundant but can be cleaned up separately.
- No changes to `stream_data_on()` / `stream_data_off()`.
- No new tests (project has no unit-test coverage for `setup()` today; adding a test scaffold is a separate effort).

Reviewers: please flag any of the above you'd like folded into this PR rather than handled as follow-ups.

- [ ] Yes
- [x] No

The exit state of `setup()` w.r.t. the streaming register changes from `1` to `0`. Any caller that runs `S.setup()` and immediately expects to be streaming will need to call `S.stream_data_on()` (the documented data-acquisition entry point), but in practice that is what callers already do — and the previously documented behavior of `setup()` was the bug being fixed.

## Test plan
- [x] `flake8 --count python/` clean (matches CI invocation)
- [x] `python -m py_compile python/pysmurf/client/base/smurf_control.py` clean
- [ ] On a live SMuRF carrier: run `S.setup()` and confirm `S.get_stream_enable() == 0` after it returns
- [ ] On a live SMuRF carrier: confirm `S.stream_data_on()` / `S.stream_data_off()` still produce a data file as before

### DIFF
--- a/python/pysmurf/client/base/smurf_control.py
+++ b/python/pysmurf/client/base/smurf_control.py
@@ -337,7 +337,10 @@ class SmurfControl(SmurfCommandMixin,
         - Turns off flux ramp, but configures based on values for
           reset rate and fraction full scale provided in pysmurf
           configuration file.
-        - Enables data streaming.
+        - Briefly enables data streaming during configuration, then
+          disables it before returning.  Callers that want to take
+          data should use ``stream_data_on()`` / ``stream_data_off()``
+          (see slaclab/pysmurf#652).
         - Sets mask and payload size to a single channel.
         - Sets RF amplifier biases (without enabling drain voltages).
         - Configures timing based on pysmurf configuration file settings.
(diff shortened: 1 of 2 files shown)


ruck314:
Closing this PR because the associated issue it addresses, #652, is labelled `wontfix`. Not pursuing this change.

PR #902, opened 2026-04-30
fix(SmurfControl): Allow loading cfg_file in offline mode (#609)

### LINKED ISSUE #609: Should be able to load cfg file when running in offline mode
## Describe the bug
Currently it is not possible to load a cfg file into pysmurf when running in offline mode. 

## To Reproduce
Try to create a pysmurf instance like:
```
S = SmurfControl(offline=True, cfg_file=<cfg_path>)
```

## Expected behavior
Pysmurf should be able to load properties from a cfg file so we can access properties like `S.pA_per_phi0` and `S.R_sh` when running analysis offline.

## Additional details
I've tested and running these two lines manually imports the config properly and has the desired behavior: https://github.com/slaclab/pysmurf/blob/5e231d4445e34095f5ab7728dccaa09978d83335/python/pysmurf/client/base/smurf_control.py#L120-L123

However specifying the cfg file when initializing also triggers the initialize function here:
https://github.com/slaclab/pysmurf/blob/5e231d4445e34095f5ab7728dccaa09978d83335/python/pysmurf/client/base/smurf_control.py#L147-L148
which crashes when not connected to a real system.

ruck314:
## Summary
- Fixes [#609](https://github.com/slaclab/pysmurf/issues/609) — `SmurfControl(offline=True, cfg_file=<path>)` previously crashed during `initialize()`.
- `which_bays()`, `get_crate_id()` and `get_slot_number()` resolve to `_caget()`, which returns `None` in offline mode; the next `for bay in self.bays` then raised `TypeError: argument of type 'NoneType' is not iterable`.
- Skip those three register reads when `self.offline` is `True` and derive `self.bays` from the config-defined `self._bands` (`bay = band // 4`) so the downstream band/bay validation stays consistent. `crate_id` / `slot_number` are set to `None`. The online path is byte-identical.

## Test plan
- [x] Reproduced the original crash on `main` with `SmurfControl(offline=True, cfg_file=<cfg>)` → `TypeError`.
- [x] Re-ran with the patch applied: construction succeeds; `S.R_sh`, `S.pA_per_phi0`, `S.bays`, `S.crate_id`, `S.slot_number` all populated as expected.
- [x] `flake8 --count python/` clean (matches CI invocation).
- [x] Client smoke import (`python -c 'import matplotlib; matplotlib.use("Agg"); import pysmurf.client'`) passes.
- [ ] Online regression covered by existing CI.

## Out of scope

- Making `setup()` offline-safe (not requested by #609; offline users don't pass `setup=True`).
- Refactoring `which_bays`/`get_crate_id`/`get_slot_number` themselves to be offline-aware — extra surface area, not needed.
- Changes to `base_class.py` — already handled by commit `16b44046`.

Reviewers: flag any of the above as in-scope and we can fold it into this PR.

### DIFF
--- a/python/pysmurf/client/base/smurf_control.py
+++ b/python/pysmurf/client/base/smurf_control.py
@@ -249,12 +249,20 @@ class SmurfControl(SmurfCommandMixin,
             else:
                 self.log.set_logfile(None)
 
-            # Which bays were enabled on pysmurf server startup?
-            self.bays = self.which_bays()
+            if not self.offline:
+                # Which bays were enabled on pysmurf server startup?
+                self.bays = self.which_bays()
 
-            # Crate/carrier configuration details that won't change.
-            self.crate_id = self.get_crate_id()
-            self.slot_number = self.get_slot_number()
+                # Crate/carrier configuration details that won't change.
+                self.crate_id = self.get_crate_id()
+                self.slot_number = self.get_slot_number()
+            else:
+                # Offline: no hardware to query. Derive bays from the
+                # config-defined bands so downstream band/bay validation
+                # is consistent.
+                self.bays = sorted({b // 4 for b in self._bands})
+                self.crate_id = None
+                self.slot_number = None
 
             # Channel assignment files
             self.channel_assignment_files = {}


ruck314:
Closing in light of Tristan's refactor proposal. Stage 4 replaces smurf_config.py / smurf_config_properties.py with a single inherited schema, and a reattaching client needs no cfg file. Leaving #609 open.

PR #907, opened 2026-04-30
fix(rtm_slow_dac): Make DAC chip parameters configurable (#543)

### LINKED ISSUE #543: Fix hard code of RTM slow DAC parameters
Here:
https://github.com/slaclab/pysmurf/blob/a370b515ab717c982781223da147bea3c8fb3a9c/python/pysmurf/client/base/base_class.py#L184

ruck314:
## Summary

Closes #543.

The RTM slow DAC parameters (`_rtm_slow_dac_max_volt`, `_rtm_slow_dac_nbits`, derived `_rtm_slow_dac_bit_to_volt`) were hardcoded in `SmurfBase.__init__` to AD5790 / 10 V reference values. RTMs using a different DAC chip or reference voltage produced silently wrong physical units in TES bias commands, IV curves, and TES bipolar waveform playback. The `±9.9 V` soft-warning thresholds in `get_rtm_slow_dac_volt` were similarly baked in.

This PR exposes `rtm_slow_dac_max_volt` and `rtm_slow_dac_nbits` through the existing `SmurfConfigPropertiesMixin` / cfg-schema pattern (same approach as `pA_per_phi0`, `hemt_bit_to_V`, `fraction_full_scale`, etc.). Defaults match the AD5790 with a 10 V reference, so existing cfg files and offline mode are unaffected.

### Changes

- **`smurf_config.py`** — new `rtm_slow_dac` schema block with AD5790 defaults and validators (`max_volt > 0`, `nbits > 0`).
- **`smurf_config_properties.py`** — three new underscore attrs in `__init__`, loader call in `copy_config_to_properties`, and `@property` getter/setters for `rtm_slow_dac_max_volt`, `rtm_slow_dac_nbits`, and the derived `rtm_slow_dac_bit_to_volt`. Setting either input recomputes `bit_to_volt` automatically.
- **`base_class.py`** — old unconditional hardcode replaced with a fallback-only block that respects values already populated by the mixin/cfg path. Offline / standalone usage still gets the AD5790 defaults bit-for-bit.
- **`smurf_command.py`** — `±9.9 V` warning thresholds in `get_rtm_slow_dac_volt` now derived as `0.99 * _rtm_slow_dac_max_volt`, so they auto-scale to the configured DAC range.
- **`cfg_files/template/template.cfg`** — commented-out documentation block. Existing site cfgs (b33, kx, lab1, nist, pton, rflab, slac, stanford, ucsd) need no changes — schema injects defaults invisibly.

### Backward compatibility

Verified bit-for-bit: with no `rtm_slow_dac` entry in the cfg, `_rtm_slow_dac_bit_to_volt` evaluates to the exact same `2*10./2**20` value as before.

## Test plan

- [x] `flake8 --count python/` (CI's invocation) — 0 errors
- [x] Schema injects AD5790 defaults when cfg lacks `rtm_slow_dac` block
- [x] Full override `{rtm_slow_dac_max_volt: 5.0, rtm_slow_dac_nbits: 16}` round-trips through schema + mixin
- [x] Partial override (only `max_volt`) keeps the other field at its default
- [x] Invalid values (`max_volt: -1`, `nbits: 0`) raise `SchemaError`
- [x] Property setter recompute: changing `rtm_slow_dac_max_volt` or `rtm_slow_dac_nbits` at runtime updates `_rtm_slow_dac_bit_to_volt` automatically
- [x] All three `SmurfBase.__init__` paths (cfg-populated mixin / mixin-with-None / standalone-no-mixin) produce correct values
- [x] `0.99 * 10.0 == 9.9` confirms warning threshold matches the old hardcode
- [ ] Hardware regression on a real RTM with default cfg (smoke test, no behavior change expected)

### DIFF
--- a/cfg_files/template/template.cfg
+++ b/cfg_files/template/template.cfg
@@ -145,6 +145,14 @@
 "bias_line_resistance" : 15800,
 "high_low_current_ratio" : 6.08,
 
+# RTM slow DAC chip parameters.  Defaults match the AD5790 with a 10V
+# reference voltage on standard RTM revisions.  Override only if your
+# RTM uses a different DAC chip or reference voltage.
+# "rtm_slow_dac": {
+#         "rtm_slow_dac_max_volt": 10.0,
+#         "rtm_slow_dac_nbits": 20
+# },
+
 "high_current_mode_bool": 0,
 
 "all_bias_groups": [7],
--- a/python/pysmurf/client/base/base_class.py
+++ b/python/pysmurf/client/base/base_class.py
@@ -235,15 +235,22 @@ class SmurfBase:
 
         self.freq_resp = {}
 
-        # RTM slow DAC parameters (used, e.g., for TES biasing). The
-        # DACs are AD5790 chips
-        self._rtm_slow_dac_max_volt = 10. # Max unipolar DAC voltage,
-                                        # in Volts
-        self._rtm_slow_dac_nbits = 20
+        # RTM slow DAC parameters (used, e.g., for TES biasing).
+        # Defaults match the AD5790 chips on standard RTMs with a
+        # 10V reference voltage.  Override via the `rtm_slow_dac`
+        # section of the pysmurf configuration file for boards that
+        # use a different DAC chip or reference voltage.  Values may
+        # have already been set by SmurfConfigPropertiesMixin from
+        # the cfg file, in which case those win.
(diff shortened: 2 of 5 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. Stage 4 replaces smurf_config.py / smurf_config_properties.py / template.cfg and the cfg_files layout with a single inherited config schema, so this belongs in the new schema rather than in code slated for deletion. Leaving #543 open.

PR #908, opened 2026-04-30
fix(SmurfCommand): Handle None pysmurf version in JESD checks and setDefaults (#538)

### LINKED ISSUE #538: Setup stall when rogue docker uses v4.0.0 on JESD health check
## Describe the bug

Ed reports a stall out on attempt to setup using rogue v4.0.0 docker on the JESD health checks:
image

Probably because the pysmurf version poll is timing out so the caget returns None, which breaks the version checking at the start of the JESD health check.

ruck314:
## Summary
- Fixes #538 — setup stalled on the JESD health check when the pysmurf version register read returned `None` (caget timeout, offline mode, or non-executing `_caget` call), crashing with `AttributeError: 'NoneType' object has no attribute 'split'`.
- Guards all three call sites of `self.get_pysmurf_version(**kwargs).split('+')[0]` in `python/pysmurf/client/command/smurf_command.py`: `set_check_jesd` (the issue's primary symptom), `get_jesd_status`, and `set_defaults_pv`.
- On `None`, each method logs an error and exits via its existing failure return — `None` for the JESD methods, `False` for `set_defaults_pv` to match its `TimeoutError` exit at line 639. No public-API signature changes.

## Test plan
- [x] `flake8 --count python/` (matches CI invocation in `.github/workflows/test-or-deploy.yml`) → 0 issues
- [x] `grep -n "get_pysmurf_version" python/pysmurf/client/command/smurf_command.py` confirms all 3 call sites are now guarded; no other call sites exist
- [ ] Hardware-in-the-loop verification of the original rogue v4.0.0 setup stall scenario (out of scope for repo CI)

### DIFF
--- a/python/pysmurf/client/command/smurf_command.py
+++ b/python/pysmurf/client/command/smurf_command.py
@@ -612,7 +612,13 @@ class SmurfCommandMixin(SmurfBase):
         """
         # strip any commit info off the end of the pysmurf version
         # string
-        pysmurf_version = self.get_pysmurf_version(**kwargs).split('+')[0]
+        pysmurf_version = self.get_pysmurf_version(**kwargs)
+        if pysmurf_version is None:
+            self.log(
+                'Could not read the pysmurf version register; skipping'
+                ' setDefaults configuration validation.', self.LOG_ERROR)
+            return False
+        pysmurf_version = pysmurf_version.split('+')[0]
 
         # Extra registers allow confirmation of successful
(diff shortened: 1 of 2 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. Stage 3 replaces the hand-written smurf_command/base_class register layer (and its docstrings) with runtime introspection and a generated shim, so this change would be discarded. Leaving #538 open to be revisited on the new layer.

PR #913, opened 2026-04-30
feat(estimate_phase_delay): Add phaseDelaySubbands config field for resonator-free latency fit (#513)

### LINKED ISSUE #513: Define a subband to measure a latency and phase offset
## Describe the problem
Define a subband without resonators to measure latency and phase offset. Save it to config file since it is HW depenent. 

## Additional details
Add any other details about the feature request here.

ruck314:
## Summary
- Adds an Optional per-band config field `phaseDelaySubbands` listing DSP subband indices known to be empty of resonators on this hardware.
- When set, `estimate_phase_delay()` uses those subbands for the DSP latency polyfit instead of geometrically selecting them from `freq_min`/`freq_max`, eliminating the resonator-induced phase distortion that would otherwise bias the linear phase-vs-frequency fit.
- Backward compatible: existing configs hit the original geometric-selection path byte-identical to pre-PR.

Closes #513.

## Changes
- `python/pysmurf/client/base/smurf_config.py` — schema entry validating `list[int]` in `[0, 128)` or `None`; explicit `null` and empty list both accepted at validation (use-time check raises if every entry is unprocessed). Adds `Or` to the schema import.
- `python/pysmurf/client/base/smurf_config_properties.py` — `phase_delay_subbands` per-band dict property, populated by `copy_config_to_properties()`.
- `python/pysmurf/client/util/smurf_util.py` — new `subband` kwarg on `estimate_phase_delay()` with precedence **arg > config (`self.phase_delay_subbands[band]`) > geometric (`freq_min`/`freq_max`)**. When an explicit list is in use, the `freq_min`/`freq_max` post-fit window is bypassed for the DSP fit only; the cable-delay fit (broadband, no resonator issue) is unaffected. Logs which subbands are in use; raises `ValueError` on empty processed-subband resolution.
- `cfg_files/template/template.cfg` — commented-out example in each of the four `band_N` blocks.
- `README.config_file.md` — documents `phaseDelaySubbands` and `bandDelayUs`; marks `refPhaseDelay`/`refPhaseDelayFine` as deprecated.

## Test plan
- [x] `flake8 --count python/` → 0 (matches CI invocation)
- [x] `python3 -m py_compile` on all three modified Python modules — clean
- [x] Schema-validation offline against `cfg_files/template/template.cfg`:
  - field absent → `None` per band (backward compat)
  - explicit `[50, 51, 52]` → accepted
  - out-of-range `[200]` → `SchemaError`
  - explicit `null` → accepted as `None`
  - empty list `[]` → accepted (use-time validation)
- [ ] On-hardware: run `S.estimate_phase_delay(band)` with `phaseDelaySubbands` populated for a band that contains resonators; confirm log line `using N user-specified resonator-free subband(s)` and verify cleaner sub-radian post-correction residuals vs. an unconfigured run.
- [ ] On-hardware backward-compat: load an existing per-site config without `phaseDelaySubbands`; confirm `S.estimate_phase_delay(band)` produces numerically identical output to pre-PR.

### DIFF
--- a/README.config_file.md
+++ b/README.config_file.md
@@ -12,7 +12,7 @@ The configuration files are stored in the `cfg_files` folder in the pysmurf git 
 4. If you know it, update the `bias_line_resistance` and `R_sh` fields.
 5. Save
 
-You should be able to call this config file when you instantiate your `SmurfControl` object. One of the first steps you'll do is run `estimate_phase_delay`. You can save the outputs into your config file so you will not need to run it in the future. 
+You should be able to call this config file when you instantiate your `SmurfControl` object. One of the first steps you'll do is run `estimate_phase_delay`. You can save the outputs into your config file so you will not need to run it in the future. For best results on detector arrays whose resonator placement is known, populate `phaseDelaySubbands` per band with subbands free of resonators before running it.
 
 # Important Variables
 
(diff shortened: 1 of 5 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. Stage 4 replaces smurf_config.py / smurf_config_properties.py / template.cfg and the cfg_files layout with a single inherited config schema, so this belongs in the new schema rather than in code slated for deletion. Leaving #513 open.

Issue #818, opened 2025-12-10
Major restructuring to support RFSoC and SPT


tristpinsm:
With the move to support RFSoC boards and the SPT 3G+ camera, we would like to restructure `pysmurf` and support libraries essentially from the ground up.

## `pysmurf` rebuild

Some ideas that have been discussed so far (to be broken down into issues as we progress):
- clearer separation between "core" functionality and instrument-specific features
  - should be reflected in a clearer separation between client and server code
  - move more core stuff into server
  - thin client, specific implementation for a given instrument built on abstract base class
- Simplify and restructure configuration and state
  - state should not be stored in configuration, move to FW
  - make configuration files hierarchical so that higher levels can inherit and we avoid duplication
- "firmware" vs "software
  - some software is currently implemented in `cryo-det`
  - should move back into `pysmurf` unless access to server-side register is needed (may be obviated by move to rogue 6)
- Simplify release scripts
  - move away from complicated and cumbersome bash scripts
  - maybe just maintain Dockerfile and docker-compose configs
  - consider distrobox.it as an alternative to docker compose
- Simplify startup scripts
  - current system is extremely hard to understand
  - argument parsing is very convoluted
  - nested roots of PCIe and cryo is problematic
    - PCIe should be a device, not a root

## SPT 3G+ support

Still defining requirements:
- First task is defining an interface document
- Minimise application-specific shims
- get SPT GPS spec

Rough timeline:
- first run in the lab (Argon, Chicago, Berkeley?)
  - includes timing, detectors, etc
- Camera deployment (28/29)
- Computing system send down season previous to deployment (27/28)
  - doesn’t need to include readout
  - GPS system
- Mid-27 timescale for in lab testing

ruck314:
https://github.com/slaclab/pysmurf/pull/914

PR #914, opened 2026-05-01
Add basic rfsoc + pre-spectra support

swh76:
## Description

Adds support for RFSoC-based hardware (ZCU208) as an alternative to the existing ATCA-crate-based systems. All changes are additive and fully backwards-compatible: the ATCA production path is unaffected when the new flags are not passed.

### New hardware support

The ZCU208 RFSoC board replaces the ATCA crate, AMC carrier, and AMC daughter cards with a single commercial RFSoC board running embedded Linux. It communicates with the server host over Ethernet via the existing PCIe offload card, so `cmb_pcie.py` is still the correct server entry point.

Two firmware variants are supported:

- `MicrowaveMuxZcu208_BaseBand` — standard RFSoC firmware, selected with `--is-rfsoc`
- `MicrowaveMuxZcu208_PreSpectra` — pre-Spectra RFSoC firmware, selected with `--is-prespectra` (implies `--is-rfsoc`)

### Firmware checking and auto-programming

When `--is-rfsoc` is passed, the server startup script (`server_common.sh`) now:

1. SSHes to the RFSoC management interface (`--rfsoc-mgmt-ip`, default `10.0.1.200`) and reads the running firmware target and git hash via `axiversiondump`
2. Selects the correct `linux.tar.gz` image from `/tmp/fw` based on the requested variant (`BaseBand` or `PreSpectra`)
3. If the hash or target doesn't match, copies `system.bit` to the board over SCP, reboots it, and verifies the new hash after reboot
4. Retries `axiversiondump` for up to 60 seconds after ping comes up, to handle the window where the board is reachable but not yet fully initialised

Firmware checking can be disabled with the existing `-D`/`--no-check-fw` flag.

### Docker image changes

`definitions.sh` is refactored from flat per-system variables to a `FW_SYSTEMS` array, with each entry specifying `name|repo_url|tag|fw_files[|zip_file]`. This makes adding or disabling a hardware system a one-line change. `validate.sh` and `build.sh` loop over this array generically. The RFSoC firmware images (`linux.tar.gz`) and pyrogue ZIP are downloaded alongside the existing ATCA assets.

### pyrogue module loading

`CmbPcie.py` previously imported `CryoDet._MicrowaveMuxBpEthGen2.FpgaTopLevel` at module level, which would fail if the pyrogue ZIP was not yet on `sys.path`. The import is now deferred to a helper function called inside `__init__`, after the ZIP has been added to `sys.path`. For RFSoC systems the import resolves to `CryoDet._MicrowaveMuxZcu208.FpgaTopLevel` (a subclass that defaults `isRFSOC=True`); for ATCA it resolves to `_MicrowaveMuxBpEthGen2` as before.

### Implied flags

To reduce error-prone manual flag specification:
- `--is-rfsoc` automatically adds `--disable-bay0` and `--disable-bay1` (RFSoC has no AMC bays)
- `--is-prespectra` implies `--is-rfsoc` and therefore also adds `--disable-bay0` and `--disable-bay1`
- When `--is-rfsoc` is active, `-d /tmp/fw/smurf_cfg/defaults/defaults_rfsoc_zcu208.yml` is passed automatically unless the user explicitly provides `-d`

### Note on TKID

The TKID firmware entry in `definitions.sh` is commented out in this branch (marked "NO LONGER BEING DEVELOPED").  We no longer plan to support this option and are dropping it to reduce deployment docker size.

## Tests done on this branch

- Full server startup with `--is-rfsoc` on a ZCU208 board: firmware hash check passes when board is up to date
- Firmware auto-programming: board running `MicrowaveMuxZcu208_BaseBand` with mismatched hash is successfully reprogrammed and verified
- Firmware target switching: board running `BaseBand` firmware correctly reprogrammed to `PreSpectra` when `--is-prespectra` is passed
- `axiversiondump` retry logic: confirmed server waits correctly when board is slow to initialise after reboot
- ATCA production path: server startup with standard ATCA flags (`-S`, `-N`, `-a`) confirmed unaffected by all changes
- Docker image build: `build_pre-spectra_local.sh` successfully validates and downloads all three system firmware entries (umux, rfsoc BaseBand, rfsoc PreSpectra) and builds the image
- `--disable-bay0`/`--disable-bay1` confirmed auto-injected when `--is-rfsoc` or `--is-prespectra` passed
- Default YML auto-selection confirmed for RFSoC path; user-supplied `-d` correctly overrides it

#### Before
```python
def __init__(self,
             ...
             disable_bay0   = False,
             disable_bay1   = False,
             enable_pwri2c  = False,
             ...):
```

#### After
```python
def __init__(self,
             ...
             disable_bay0   = False,
             disable_bay1   = False,
             is_rfsoc       = False,
             is_prespectra  = False,
             enable_pwri2c  = False,
             ...):
```

Two new optional keyword arguments added, both defaulting to `False`. Existing callers are unaffected.

---

#### Before
```python
def __init__(self,
             ...
             disable_bay0   = False,
             disable_bay1   = False,
             enable_pwri2c  = False,
             ...):
```

#### After
```python
def __init__(self,
             ...
             disable_bay0   = False,
             disable_bay1   = False,
             is_rfsoc       = False,
             is_prespectra  = False,
             enable_pwri2c  = False,
             ...):
```

Same as `CmbPcie`. Two new optional kwargs, both defaulting to `False`.

---

#### Before
```python
def __init__(self,
             ...
             disable_bay0   = False,
             disable_bay1   = False,
             **kwargs):
```

#### After
```python
def __init__(self,
             ...
             disable_bay0   = False,
             disable_bay1   = False,
             is_rfsoc       = False,
             is_prespectra  = False,
             **kwargs):
```

Two new optional kwargs accepted and stored. No logic changes.

---

#### Before
No `--is-rfsoc`, `--rfsoc-mgmt-ip`, or `--is-prespectra` arguments existed.

#### After
Three new optional CLI arguments added to the hardware group:

| Argument | Type | Default | Description |
|---|---|---|---|
| `--is-rfsoc` | flag | `False` | Selects RFSoC hardware path |
| `--rfsoc-mgmt-ip` | string | `10.0.1.200` | RFSoC embedded-processor management IP |
| `--is-prespectra` | flag | `False` | Selects pre-Spectra RFSoC firmware variant; implies `--is-rfsoc` |

---

#### Before
`arg_parser` did not handle `--is-rfsoc`, `--is-prespectra`, `--rfsoc-mgmt-ip`, or `-d`/`--defaults`. These fell through to the catch-all `*` case and were passed to the pysmurf server verbatim as separate tokens, which caused `-d <path>` to be split across two iterations.

#### After
All four are now explicitly handled. `-d`/`--defaults` is consumed as a key-value pair (fixing the split-token bug). `--is-rfsoc` and `--is-prespectra` trigger automatic injection of `--disable-bay0`, `--disable-bay1`, and (for pre-spectra) `--is-prespectra` into the pysmurf server args. `initialize()` now branches on `is_rfsoc`: the ATCA code path is completely unchanged; the RFSoC path calls `checkRFSoCFW`, `findPyrogueFiles("rfsoc")`, and auto-selects the defaults file.

---

#### Before
Firmware was defined as flat shell variables per system:
```bash
fw_repo=...
fw_repo_tag=...
mcs_file_name=...
tkid_fw_repo=...
tkid_fw_repo_tag=...
tkid_mcs_file_name=...
```

#### After
Firmware is defined as a bash array with one pipe-separated entry per system:
```bash
FW_SYSTEMS=(
    "name|repo_url|tag|fw_files[|zip_file]"
    ...
)
```
`validate.sh` and `build.sh` loop over this array generically. To add a new system: append one line. To disable a system: comment out one line.

https://github.com/slaclab/pysmurf/issues/818


python/pysmurf/core/roots/CmbEth.py
tristpinsm, line 26: I guess there is no RFSoC configuration over ethernet (or the PCIe version is ethernet), so this class should never be called with `is_rfsoc=True` or `is_prespectra=True` ? If that's true, maybe it should raise an exception if those set to True?

python/pysmurf/core/roots/Common.py
```diff
@@ -42,6 +42,8 @@ def __init__(self, *,
                  pcie           = None,
                  disable_bay0   = False,
                  disable_bay1   = False,
+                 is_rfsoc       = False,
```
tristpinsm: not really important, but these arguments appear to be unused and could be removed -- also where they are passed to `Common.__init__` in the child classes.

server_scripts/dev_board_eth.py
```diff
@@ -58,6 +58,8 @@
                            pv_dump_file   = args['pv_dump_file'],
                            disable_bay0   = args['disable_bay0'],
                            disable_bay1   = args['disable_bay1'],
+                           is_rfsoc       = args['is_rfsoc'],
+                           is_prespectra  = args['is_prespectra],
```
tristpinsm:
```suggestion
                           is_prespectra  = args['is_prespectra'],
```

tristpinsm:
I will double-check that starting up ATCA SMuRF with this branch still works.

Outside the scope of this PR, but when we start thinking about re-factoring `pysmurf`, building in modularity so that the startup sequence is simpler and support more platforms more easily would be nice.

Also avoiding commiting notebooks and scratch stuff in that potential future.

docker/server/validate.sh
```diff
@@ -12,111 +11,90 @@ function error {
...
-    else
-	    echo "Found."
-    fi
+    local page=1
+
+    echo "Searching for tag ${tag} in repo $1 ..."
+
+    while true; do
+        local api_url="${api_base}?per_page=100&page=${page}"
+        local tags=$(curl \
```
tristpinsm:
```suggestion
        local tags
        tags=$(curl \
          --silent \
          --url "${api_url}" \
          --header "Authorization: token ${GITHUB_TOKEN}" \
          --fail)
        rc=$?

        if [ $rc -ne 0 ]; then
```
My token doesn't have access to the the Zcu repo (I do, I didn't set up that token for it), but this loop wasn't failing, it was going forever. Turns out the status code was being captured for the local variable assignment.

tristpinsm:
ok, with a few more small fixes I'm able to program an ATCA smurf board and run `setup` to completion

tristpinsm (approved):

PR #915, opened 2026-05-01
fix(find_freq): Warn user when ADC rails during amplitude sweep (#511)

### LINKED ISSUE #511: User should be warned if find_freq rails ADC on any subband.
Just ran into this issue at Princeton ; the `find_freq` output is very strange if it rails the ADC.  Need to think about the best way to implement without increasing how long `find_freq` takes to execute since it's already pretty slow.

ruck314:
## Summary
- Closes #511. Adds a post-scan ADC-railing warning to `find_freq` so users know when the amplitude sweep produced bad data because the ADC saturated.
- After `full_band_ampl_sweep` returns, the function inspects each requested subband's normalized response magnitude and logs a red `LOG_ERROR` message naming any subband whose `|resp|` crosses `0.95` — the symptom shown in the issue's screenshot. The list is also stashed in `self.freq_resp[band]['find_freq']['railed_subbands']` for downstream inspection/tests.
- Gated by a new `check_if_adc_is_saturated=True` kwarg, matching the pattern `full_band_resp` already uses (smurf_tune.py:585). Warns rather than raises so partial scan data remains usable for resonators in unrailed subbands.
- Zero added firmware time: detection reuses the response data already collected by the sweep, addressing the issue author's concern that `find_freq` is "already pretty slow."

## Out of scope (intentionally)
- **Adding a firmware register that reports saturation during the eta scan** (suggested in the issue thread by @swh76). That's a Rogue/firmware change and outside this Python-only fix.
- **Adding the same check to `fast_eta_scan` directly.** It returns raw I/Q with no normalization context guaranteed by all callers; doing it at `find_freq` keeps the surface minimal and matches the issue title.
- **Auto-adjusting attenuation.** The issue asks for a warning, not auto-correction.

Reviewers — please flag if any of the above should be folded into this PR rather than deferred.

## Test plan
- [x] `flake8 --count python/` → 0 errors (matches CI invocation in `.github/workflows/test-or-deploy.yml`).
- [x] AST parse of the modified file passes.
- [x] Offline numpy sanity check of the threshold logic with railed / clean / just-under-threshold synthetic subbands picks the correct subband.
- [ ] Hardware regression on a SMuRF system: run `S.find_freq(band, ...)` at deliberately too-low DC attenuation, confirm the red warning fires and `S.freq_resp[band]['find_freq']['railed_subbands']` is non-empty; re-run at nominal attenuation and confirm the list is empty and no warning is emitted.

### DIFF
--- a/python/pysmurf/client/tune/smurf_tune.py
+++ b/python/pysmurf/client/tune/smurf_tune.py
@@ -3440,7 +3440,8 @@ class SmurfTuneMixin(SmurfBase):
             plotname_append='', window=50, rolling_med=True,
             make_subband_plot=False, show_plot=False, grad_cut=.05,
             flip_phase=False, grad_kernel_width=8,
-            amp_cut=.25, pad=2, min_gap=2):
+            amp_cut=.25, pad=2, min_gap=2,
+            check_if_adc_is_saturated=True):
         '''
         Finds the resonances in a band (and specified subbands)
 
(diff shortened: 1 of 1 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. This adds new functionality to the SmurfControl mixins, which stage 6 dissolves into functions over a session, and the proposal keeps every stage net-negative in client surface. Better re-proposed on the new session API once stage 2 settles the interface. Leaving #511 open.

PR #917, opened 2026-05-01
feat(run_serial_*): Add use_tracked_freq option for biased-TES retune (#736)

### LINKED ISSUE #736: Serial Eta Scan should use starting freq from tracking
## Describe the problem
Currently, gradient descent and eta scan functions use the resonance frequencies in the `center_freq_array` variable as their starting point. If the TESs are biased, then the resonator could be shifted by ~100 kHz depending on phase of the SQUID induced by the TES bias current. This means that these functions _only_ work when the detectors are unbiased, which is problematic if we want to track the phase of the cable and system as the 300K temp varies without rebiasing detectors.

Instead of using the `center_freq_array` values as the starting points, these functions should use the actual tracked frequency as the starting point, since this will more accurately reflect the frequency of the resonator. I think this can be acquired through take_debug_data before calling the grad-descent and eta-scan procedures.

ruck314:
## Summary
Closes #736.

`run_serial_gradient_descent` and `run_serial_eta_scan` start their search at `centerFrequencyMHz`, which `setup_notches` writes once from the initial `find_freq` sweep. When TESs are biased, SQUID phase shifts each resonator by ~100 kHz, so the firmware capture window misses the new minimum and re-tuning is impossible without first unbiasing the detectors.

This adds `SmurfUtilMixin.update_centers_from_tracked_freq(band, nsamp=2**18, write_log=False)`, which takes one `take_debug_data` acquisition with `IQstream=0, single_channel_readout=0` while the flux ramp is running, averages the per-sample tracked-frequency offset, and writes the bias-shifted values back to `centerFrequencyMHz` for every channel in `which_on(band)`. It restores `setIQStreamEnable=1` on exit, matching the convention at the tail of `tracking_setup`. Channels not in `which_on` are left untouched. Returns the full 512-element array as written.

Both `run_serial_gradient_descent` and `run_serial_eta_scan` gain an opt-in `use_tracked_freq=False, nsamp=2**18` kwarg pair. When `use_tracked_freq=True`, the helper is called **before** `flux_ramp_off()` (tracked frequency is only meaningful while tracking is active). Default-False preserves legacy behavior — the existing call sites in `setup_notches` (`smurf_tune.py:402,406`), `noise_vs_bias` (`smurf_noise.py:668-669`), and `take_noise_high_bandwidth` (`smurf_noise.py:1992-1994`) are byte-for-byte equivalent.

Diff is +128/-4 across two files, no new imports.

## Test plan
- [x] `flake8 --count python/` reports 0 issues (matches CI invocation in `.github/workflows/test-or-deploy.yml`)
- [x] `python -c \"import ast, pathlib; ast.parse(pathlib.Path('python/pysmurf/client/util/smurf_util.py').read_text()); ast.parse(pathlib.Path('python/pysmurf/client/command/smurf_command.py').read_text())\"` parses cleanly
- [x] AST signature check confirms `use_tracked_freq` and `nsamp` kwargs added with default-False on both `run_serial_*` functions, helper present on `SmurfUtilMixin`
- [x] `grep -n run_serial_{gradient_descent,eta_scan} python/pysmurf/client/{tune,debug}/*.py` confirms internal callers unchanged (default-False legacy path)
- [ ] Hardware integration: tune a band, run `tracking_setup` with the flux ramp running, apply a TES bias known to shift resonators, then:
  ```python
  # Baseline (existing bug): expect convergence failure / wrong minimum on biased channels
  S.run_serial_gradient_descent(band)
  # Fix: expect convergence on the same biased channels
  S.run_serial_gradient_descent(band, use_tracked_freq=True)
  ```
- [ ] Hardware integration: verify `S.get_center_frequency_array(band)` after the fix call shifts only `which_on` channels by ~the SQUID phase shift; non-on channels unchanged
- [ ] Hardware integration: verify `S.get_iq_stream_enable(band) == 1` after the helper returns (regression check on the restore)
- [ ] Hardware integration: repeat the above with `S.run_serial_eta_scan(band, use_tracked_freq=True)`
- [ ] Hardware integration (empty band): with no channels on, call `S.run_serial_gradient_descent(band, use_tracked_freq=True)`; expect a single log line and `centerFrequencyMHz` unchanged

### DIFF
--- a/python/pysmurf/client/command/smurf_command.py
+++ b/python/pysmurf/client/command/smurf_command.py
@@ -828,7 +828,8 @@ class SmurfCommandMixin(SmurfBase):
 
     _run_serial_eta_scan_reg = 'runSerialEtaScan'
 
-    def run_serial_eta_scan(self, band, timeout=240, **kwargs):
+    def run_serial_eta_scan(self, band, timeout=240,
+            use_tracked_freq=False, nsamp=2**18, **kwargs):
         """
         Does an eta scan serially across the entire band. You must
         already be tuned close to the resontor dip. Use
@@ -840,7 +841,24 @@ class SmurfCommandMixin(SmurfBase):
             The band to eta scan.
         timeout : float, optional, default 240
             The maximum amount of time to wait for the PV.
+        use_tracked_freq : bool, optional, default False
+            If True, calls
+            :func:`~pysmurf.client.util.smurf_util.SmurfUtilMixin.update_centers_from_tracked_freq`
+            before turning the flux ramp off, so the scan starts from
+            the bias-shifted resonator frequencies measured by the
+            tracking algorithm rather than the originally programmed
+            centers. Requires the flux ramp to be running and tracking
+            to be set up. Default False preserves legacy behavior.
+        nsamp : int, optional, default 2**18
+            Number of samples used by
+            ``update_centers_from_tracked_freq`` when
+            ``use_tracked_freq`` is True. Ignored otherwise.
         """
+
+        if use_tracked_freq:
+            self.update_centers_from_tracked_freq(
+                band, nsamp=nsamp,
+                write_log=kwargs.get('write_log', False))
 
         # need flux ramp off for this - enforce
         self.flux_ramp_off()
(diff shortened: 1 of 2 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. This adds new functionality to the SmurfControl mixins, which stage 6 dissolves into functions over a session, and the proposal keeps every stage net-negative in client surface. Better re-proposed on the new session API once stage 2 settles the interface. Leaving #736 open.

PR #918, opened 2026-05-01
fix(assign_channels): Try next-closest subband when closest is full (#658)

### LINKED ISSUE #658: Do not blindly assign resonances to same subband
## Describe the problem
This issue references the change to one channel per subband with 512 subbands.

If two resonators are spaced by less than half the width of a subband (<1.2 MHz), they could be placed within the same subband, with the higher-frequency resonance then not receiving a channel assignment.
There is not an opportunity to assign a different subband or choose which of the resonators does not get a channel assignment.

Below is an illustration of this occuring in a SO v3.2 umux chip. Shown is a find_freq measurement overplotted with the locations of resonators which have been labelled by their corresponding umux chip bondpad number.
The two resonances on the left were both assigned to the red subband, with res56 receving a channel assignment and res60 not, and the same situation with the two resonances on the right.

image

This occured three times on this one umux chip, with resonator spacings of 0.76, 0.5, and 0.76 MHz. Another pair (not shown) had a spacing of 0.68MHz, and they were both assigned channels as a result of how they happened to line up with the subband boundaries.

Could the subband assignment be done in different manner, e.g. attempt to place it in the subband to which it is closest, but if a resonator has already been assigned to that subband, then choose the next closest subband?
For example, in the plot above res60 could be assigned to the blue subband, which is otherwise unoccupied.

At a minimum, it would be nice to be able to choose which resonance gets a channel assignment and which gets dropped, instead of always dropping the higher-frequency resonance.

(I am assuming that the switch to 512 subbands was done for other important reasons, and therefore a reversion to 128 subbands would not be considered.)

## Additional details
With the current umux chips in SO, this is leading to a few percent yield hit that did not exist in the 4 channels per subband, 128 subband version of the s/w and f/w.
The resulting (soft) requirement on minimum resonator spacing of ~1 MHz was an unexpected change for NIST fab, and one they are not happy about.

ruck314:
## Summary

Closes #658.

When firmware switched to 1 channel per subband / 512 subbands, two resonators spaced less than half a subband (~1.2 MHz) apart land in the same subband and the higher-frequency one gets silently dropped. The reporter sees several-percent yield loss on SO umux chips and an unwelcome ~1 MHz minimum-spacing floor on NIST fab.

This PR makes `assign_channels()` (`new_master_assignment=True` path) fall back to the next-closest subband when the preferred one is full, up to a bounded search radius. Two-pass "assign-then-prune" is replaced with a single greedy pass that tracks per-subband usage as it iterates resonances in sorted-frequency order.

- New kwarg: `max_alt_subbands : int = 5` — bounds how far a resonator can drift from its preferred subband when collisions occur.
- All five existing call sites are unaffected (additive default kwarg).
- Behavior is byte-equivalent to the old algorithm whenever no collisions exist.
- The `min_offset` proximity prune (resonances < 0.1 MHz apart) and the `not new_master_assignment` master-list lookup path are unchanged.

## Verification

- `flake8 --count python/` → 0 (CI lint).
- Offline import smoke test passes (`python3 -c 'import pysmurf.client'`).
- 6 offline unit tests covering: byte-equivalence to old algorithm with no collisions, the exact #658 reproducer (two resonators ~0.5 MHz apart now both placed in adjacent subbands), the old algorithm dropping the higher-freq one in the same scenario, the close-pair (< 0.1 MHz) proximity prune still triggering, old-firmware 4 ch/sb path unchanged, and overflow beyond `max_alt_subbands` correctly leaving excess as `-1`.

## Out of scope

- Reverting to 128 subbands (explicitly rejected by reporter).
- Optimal global assignment (Hungarian-style). The greedy fallback is sufficient for the failure modes seen in the field; can be revisited if real-world data shows greedy is too lossy.
- The `not new_master_assignment` master-list lookup path. The reporter's complaint is about fresh assignments; lookups by definition reuse a prior assignment.

## Test plan

- [ ] Reporter (@dpdutcher) re-runs `setup_notches` with `new_master_assignment=True` on the umux chip from #658 (the one with 0.5 / 0.68 / 0.76 / 0.76 MHz pairs) and confirms all four paired resonators now receive channel assignments.
- [ ] Confirm no regression in tune/relock on a chip that previously had no collisions (assignments should be identical to pre-fix output).

### DIFF
--- a/python/pysmurf/client/tune/smurf_tune.py
+++ b/python/pysmurf/client/tune/smurf_tune.py
@@ -1634,9 +1634,16 @@ class SmurfTuneMixin(SmurfBase):
     @set_action()
     def assign_channels(self, freq, band=None, bandcenter=None,
             channel_per_subband=4, as_offset=True, min_offset=0.1,
-            new_master_assignment=False):
+            new_master_assignment=False, max_alt_subbands=5):
         """
         Figures out the subbands and channels to assign to resonators
+
+        When ``new_master_assignment=True``, each resonator is placed in
+        its closest subband with available capacity.  If the closest
+        subband is already full, the algorithm falls back to the next
+        closest, up to ``max_alt_subbands`` candidates total.  This
+        prevents silently dropping the higher-frequency resonator when
+        two resonances land in the same subband (issue #658).
 
         Args
         ----
@@ -1655,6 +1662,11 @@ class SmurfTuneMixin(SmurfBase):
         min_offset : float, optional, default 0.1
             The minimum offset between two resonators in MHz.  If
             closer, then both are ignored.
+        max_alt_subbands : int, optional, default 5
+            Maximum number of subbands to try when the closest one is
+            full.  Bounds how far a resonator can drift from its
+            preferred subband when collisions occur.  Only used when
+            ``new_master_assignment=True``.
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. The fallback adds a new max_alt_subbands parameter and search logic to assign_channels, which stage 6 redesigns as assign_channels(sess, band, freqs), so the new surface belongs there. Leaving #658 open.

PR #919, opened 2026-05-01
feat(run_iv): Add cool_voltage option for cool-wait bias (#563)

### LINKED ISSUE #563: run_iv cool_wait at a voltage variable value instead of the highest bias_voltage
## Describe the problem
The current run_iv has cool_wait at np.max(bias).

 Can there be a variable like coolbias_voltage, which if None things do the default thing and if a value, then waits there instead?

## Additional details
Waiting at different value (lower than the highest bias voltage), can help reduce the heat generated during the cool time

ruck314:
## Summary

Closes #563.

`run_iv` previously hardcoded the post-overbias cool-wait TES bias to `np.max(bias)`. This adds an optional `cool_voltage` kwarg so the caller can hold the TES at a lower voltage during the `cool_wait` window to reduce heat dissipated during cooling. When `cool_voltage=None` (default), behavior is unchanged.

- `python/pysmurf/client/debug/smurf_iv.py` — new kwarg threaded into the `overbias_tes_all` call as `tes_bias=(np.max(bias) if cool_voltage is None else cool_voltage)`, plus a matching docstring entry.

## Out of scope

- Adding `cool_voltage` to `smurf_cmd.py` CLI flags.
- Adding a config-file property — keeping parity with `cool_wait` / `overbias_*` which are kwargs only.
- Touching `overbias_tes` / `overbias_tes_all` semantics.

Reviewers: if any of the above should land in this PR rather than a follow-up, please flag it.

## Test plan

- [x] `flake8 --count python/` — 0 issues (matches `.github/workflows/test-or-deploy.yml`)
- [x] Default path: `run_iv(...)` without `cool_voltage` produces the previous `overbias_tes_all(..., tes_bias=np.max(bias), ...)` call (verified by inspection)
- [x] Override path: `run_iv(..., cool_voltage=1.0)` with `bias_high=4` calls `overbias_tes_all(..., tes_bias=1.0, ...)` (verified by inspection)
- [x] Existing callers in `python/pysmurf/client/command/smurf_cmd.py:418,425` use kwargs only, so adding the new optional kwarg is backwards-compatible
- [ ] Smoke-test on hardware: `S.run_iv(..., cool_voltage=<lower-than-bias_high>)` and confirm the TES holds at the requested voltage during `cool_wait`

### DIFF
--- a/python/pysmurf/client/debug/smurf_iv.py
+++ b/python/pysmurf/client/debug/smurf_iv.py
@@ -30,6 +30,7 @@ class SmurfIVMixin(SmurfBase):
     def run_iv(self, bias_groups=None, wait_time=.1, bias=None,
                bias_high=1.5, bias_low=0, bias_step=.005,
                show_plot=False, overbias_wait=2., cool_wait=30,
+               cool_voltage=None,
                make_plot=True, save_plot=True, plotname_append='',
                channels=None, band=None, high_current_mode=True,
                overbias_voltage=8., grid_on=True,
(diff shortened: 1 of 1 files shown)


ruck314:
Closing this PR because the associated issue it addresses, #563, is labelled `wontfix`. Not pursuing this change.

PR #920, opened 2026-05-01
feat(SmurfControl): Add ac_dc_mode cfg option for flux ramp coupling (#510)

### LINKED ISSUE #510: User should be able to specify AC/DC flux ramp mode in their experiment.cfg file.
What the title says.

ruck314:
## Summary

Closes #510.

Lets users specify the cryocard flux ramp AC/DC coupling mode from `experiment.cfg`. The hardware control (`CryoCard.set_ac_dc_relay`, added in #706) already existed but was never wired into `setup()` — the relay state was whatever the cryocard powered up with.

- `python/pysmurf/client/base/smurf_config.py` — add `Optional('ac_dc_mode', default=None)` to the `flux_ramp` schema, validated against `"AC"` / `"DC"` (case-insensitive).
- `python/pysmurf/client/base/smurf_config_properties.py` — load and expose as `flux_ramp_ac_dc_mode` property (uppercases on set; `None` when the key is absent).
- `python/pysmurf/client/base/smurf_control.py` — in `setup()`, after the existing cryocard try/except, drive the relay if the user supplied a value. Inside its own `try/except` so it does not abort `setup()` on systems with no cryocard.
- `cfg_files/template/template.cfg` — commented example documenting the new key.

When the cfg key is **absent**, `setup()` does not touch the relay, so every existing cfg file continues to behave exactly as before.

## Example usage

```jsonc
"flux_ramp" : {
    "num_flux_ramp_counter_bits" : 32,
    "ac_dc_mode" : "AC"   // or "DC"
},
```

## Out of scope

- The `select_ramp` cfg key still appears in some cfg files but is not in the schema and is not loaded at runtime. Orthogonal to #510 — left untouched.
- `flux_ramp_setup()` in `smurf_util.py` — AC/DC is a one-time hardware-init concern, not a per-ramp parameter.

## Test plan

- [x] `flake8 --count python/` — 0 findings (matches `.github/workflows/test-or-deploy.yml`)
- [x] Schema round-trip:
  - Cfg with no `ac_dc_mode` validates and `S.flux_ramp_ac_dc_mode is None`
  - `"AC"` / `"DC"` accepted; mixed case (`"Dc"`) accepted (case-insensitive)
  - `"bogus"` rejected as `SchemaError`
- [x] `cfg_files/template/template.cfg` still parses cleanly through the comment-stripping JSON loader
- [ ] Hardware smoke test on a system with a real C02/C04 cryocard:
  - Run `setup()` with `"ac_dc_mode": "AC"`, then `S.get_cryo_card_ac_dc_mode()` returns `"AC"`
  - Run again with `"DC"`, confirm `S.get_cryo_card_ac_dc_mode()` returns `"DC"`
  - With key absent, confirm no AC/DC log line and no relay write

### DIFF
--- a/cfg_files/template/template.cfg
+++ b/cfg_files/template/template.cfg
@@ -250,6 +250,10 @@
 "flux_ramp" : {
 	"select_ramp" : 1,
 	"num_flux_ramp_counter_bits" : 32
+	# "ac_dc_mode" : "DC"  # Optional. "AC" or "DC". Sets the cryocard
+	                       # flux ramp coupling relay during setup().
+	                       # If omitted, the relay is left in its
+	                       # current state. See issue #510.
 },
 
 "constant" : {
--- a/python/pysmurf/client/base/smurf_config.py
+++ b/python/pysmurf/client/base/smurf_config.py
@@ -729,7 +729,12 @@ class SmurfConfig:
         #### Start specifying flux ramp-related
         schema_dict["flux_ramp"] = {
             # 20 bits for the C0 RTMs, 32 bits for the C1 RTMs.
-            "num_flux_ramp_counter_bits" : And(int, lambda n: n in (20, 32))
+            "num_flux_ramp_counter_bits" : And(int, lambda n: n in (20, 32)),
+            # Cryocard flux-ramp coupling: "AC" or "DC". Case-insensitive.
+            # If omitted, setup() leaves the relay in whatever state the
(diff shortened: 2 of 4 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. Stage 4 replaces smurf_config.py / smurf_config_properties.py / template.cfg and the cfg_files layout with a single inherited config schema, so this belongs in the new schema rather than in code slated for deletion. Leaving #510 open.

PR #909, opened 2026-04-30
feat(SmurfUtilMixin.health_check): Aggregate system sanity checks (#528)

ruck314:
## Summary
- Closes #528 — adds a single `SmurfUtilMixin.health_check(...)` method that aggregates the existing per-component diagnostics into one consolidated sanity-check.
- Composition only; no new low-level register accessors. Reuses `get_fpga_status` (`smurf_util.py:2326`), `get_fpga_temp` (`smurf_command.py:5103`), `check_jesd` (`smurf_util.py:2263`), `CryoCard.get_fw_version` (`cryo_card.py:267`), `get_cryo_card_temp` (`smurf_command.py:5216`), `get_amplifier_biases` (`smurf_command.py:4658`), and (opt-in) `check_dac_saturation` / `check_adc_saturation` (`smurf_util.py:1658, 1686`) plus `get_carrier_sn` / `get_rtm_sn` / `get_amc_sn` (`smurf_atca_monitor.py:309/409/206`).
- Read-mostly defaults: `check_saturation=False` (calls reconfigure the DAQ MUX) and `use_atca_monitor=False` (requires the atca_monitor Rogue server) keep `health_check` safe to run mid-acquisition. Set either flag to opt in.
- Returns a structured dict (`{'ok', 'fpga', 'jesd', 'cryocard', 'saturation', 'atca'}`) and emits a colored, one-line-per-component summary via `self.log(..., self.LOG_USER)`.
- The firmware `analysisOverflow` / `synthesisOverflow` flags requested in the issue are not currently exposed by pysmurf (no Rogue YAML accessor exists). Captured in the new docstring as a follow-up to extend once the registers are plumbed.

## Test plan
- [x] `flake8 --count python/` (matches CI invocation in `.github/workflows/test-or-deploy.yml`) → 0 issues
- [x] `python -m py_compile python/pysmurf/client/util/smurf_util.py` → clean
- [x] AST verification confirms `SmurfUtilMixin.health_check` exists with signature `(self, bands=None, bays=None, check_saturation=False, use_atca_monitor=False, jesd_max_timeout_sec=60)` and `@set_action()` decorator
- [ ] Hardware-in-the-loop verification on a live carrier — out of scope for repo CI; recommended before merge:
  - `S.health_check()` → fast read-only path; confirm it returns `ok=True` on a healthy crate
  - `S.health_check(check_saturation=True)` → confirm DAC/ADC saturation per band runs cleanly
[...]

Issue #528, opened 2020-10-21
Need a pysmurf "sanity check" function


swh76:
Need a function which sanity checks the system state.  E.g.;
* If DAC is saturated for any band.
* If ADC is saturated for any band.
* If cryocard is connected.
* If JESD is up.
...what else?

swh76:
Max says they've been working on a similar script for sodetlib;

https://github.com/simonsobs/sodetlib/blob/c91671a1e5b545cef662eb05e8dce25dc2ce4c18/sodetlib/smurf_funcs/health_check.py#L70

swh76:
Should also check AMC, RTM, and carrier handles, and all the other enables/hardware ok/etc. IPMI low level regs (if the atca_monitor servers is up);
https://jira.slac.stanford.edu/browse/ESCRYODET-421

aer7:
SO has health_check!

https://github.com/slaclab/pysmurf/issues/653

PR #922, opened 2026-05-01
feat(SmurfUtilMixin.get_versions): Aggregate pysmurf/rogue/firmware versions (#492)

### LINKED ISSUE #492: Need a routine that quickly dumps the versions of *everything* from pysmurf at once.
Need a way to get configuration snapshot quickly.  E.g. what Rogue version, what pysmurf version, what dockers, etc..

ruck314:
## Summary

Closes #492 — adds a single `SmurfUtilMixin.get_versions()` aggregator so users can capture a configuration snapshot in one shot instead of calling ~7 getters and parsing `-z` out of `get_smurf_startup_args()` by hand (the recipe @swh76 sketched in the issue thread).

Composition only; no new low-level register accessors. Reuses `get_pysmurf_version` / `get_pysmurf_directory` / `get_smurf_startup_script` / `get_smurf_startup_args` / `get_enabled_bays` / `get_system_configured` / `get_rogue_version` / `get_fpga_version` / `get_fpga_uptime` / `get_fpga_git_hash[_short]` / `get_fpga_build_stamp` (all in `smurf_command.py`) and `CryoCard.get_fw_version` (`cryo_card.py:267`, accessed via `self.C` and gated on `self.offline`).

Returns a structured dict:

```python
{
    'pysmurf':  {'client_version', 'server_version', 'directory'},
    'rogue':    {'version'},
    'fpga':     {'version', 'uptime', 'git_hash', 'git_hash_short', 'build_stamp'},
    'startup':  {'script', 'args', 'zipfile'},
    'cryocard': {'fw_version'},
    'system':   {'enabled_bays', 'system_configured'},
}
```

Two notable adds over the hand-rolled snippet in the issue thread:

- Surfaces both `pysmurf.__version__` (this Python process) and `get_pysmurf_version()` (the EPICS PV the server publishes) as separate keys, so client/server skew is detectable — exactly the kind of misconfiguration the issue is trying to make visible.
- `startup.zipfile` is the basename of the rogue firmware bundle (the "docker" identifier from the issue), parsed out of the startup args via `shlex.split` and accepting `-z`, `--zip`, and `--zip=<path>` (per `python/pysmurf/core/server_scripts/Common.py:163`).

Each underlying call is wrapped in its own `try/except` so one failed PV (offline mode, server timeout) records `None` instead of aborting the whole dump. `cryocard.fw_version` is skipped when `self.offline=True` (`self.C` is a `_DummyClient` in offline mode per `base_class.py:224`, and `get_fw_version` does a real SPI read).

After populating the dict a one-line-per-component summary is emitted via `self.log(..., self.LOG_USER)`, mirroring `get_fpga_status` (`smurf_util.py:2326`).

Out of scope here (separate concerns):

- FPGA temperature / JESD lock / amplifier biases — owned by #909's `health_check` for #528.
- File-output mode — the issue says "quickly" and the example usage is interactive. If a persistent format is later wanted it's a clean follow-up.

## Test plan

- [x] `flake8 --count python/` (matches CI invocation in `.github/workflows/test-or-deploy.yml`) → `0`
- [x] `python -m py_compile python/pysmurf/client/util/smurf_util.py` → clean
- [x] AST verification confirms `SmurfUtilMixin.get_versions(self)` plus the `_extract_zipfile_from_args` static helper and `_log_versions_summary` private formatter exist with the expected signatures
- [x] `_extract_zipfile_from_args` smoke-tested for `-z <path>`, `--zip <path>`, `--zip=<path>`, missing token, empty string, and `None`
- [ ] Hardware-in-the-loop verification on a live carrier — out of scope for repo CI; recommended before merge:
  - `S.get_versions()` → confirm all six top-level keys populate and the printed summary is one line per component
  - `S.get_versions()['startup']['zipfile']` reproduces `os.path.basename` applied to the `-z` token in `S.get_smurf_startup_args()`
  - `S.get_versions()['pysmurf']['client_version']` equals `pysmurf.__version__`; `['server_version']` equals `S.get_pysmurf_version()`
  - With `S = SmurfControl(offline=True)`: returns without raising; `cryocard.fw_version` is `None`; missing PVs log warnings but don't crash

### DIFF
--- a/python/pysmurf/client/util/smurf_util.py
+++ b/python/pysmurf/client/util/smurf_util.py
@@ -16,6 +16,7 @@
 from contextlib import contextmanager
 import glob
 import os
+import shlex
 import threading
 import time
 import re
@@ -24,6 +25,7 @@ import matplotlib.pyplot as plt
 import numpy as np
 from scipy import signal
 
+import pysmurf
 from pysmurf.client.base import SmurfBase
 from pysmurf.client.command.sync_group import SyncGroup as SyncGroup
 from pysmurf.client.util.SmurfFileReader import SmurfStreamReader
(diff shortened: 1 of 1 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. Version, health and crate information become server-published self-description and platform-layer capabilities, superseding a client-side aggregator. Leaving #492 open.

Issue #486, opened 2020-08-04
Fix several issues in the `CryoCard` class.


jesusvasquez333:
## Describe the bug
While working on testing the new Cryo card firmware, I noticed several errors in the `CryoCard` class defined [here](https://github.com/slaclab/pysmurf/blob/main/python/pysmurf/client/command/cryo_card.py). 

Here is a list of the issue I found, in not particular order:
- There is an extra return statement which is never reach [here](https://github.com/slaclab/pysmurf/blob/main/python/pysmurf/client/command/cryo_card.py#L62). Seems like a remaining line from a code update.
- There is a wrong bit mask [here](https://github.com/slaclab/pysmurf/blob/main/python/pysmurf/client/command/cryo_card.py#L164). The 32-bit command structure is: 1 bit for the read flag, 11 bit for the address, and 20 bits for the data. So, to extract the address part of that 32-bit word the code should be: `return((data & 0x7FF00000) >> 20)` (note the difference in the bit mask).
- Currently, the `do_read` method returns silently when error occur [here](https://github.com/slaclab/pysmurf/blob/main/python/pysmurf/client/command/cryo_card.py#L60). It just returns zero, which is a valid return value. So at the moment there is no way to determine if this command returned a valid value. I would recommend to change that `return(0)` statement for an exception, something like this: `raise Exception(f'Error reading SPI bus. Could not read address "{address}" after {self.retry} retries')`, or something similar.  
- I would recommend that these global low level conversion methods defined [here](https://github.com/slaclab/pysmurf/blob/main/python/pysmurf/client/command/cryo_card.py#L160-L170) should be included as hidden members of the `CryoCard` class, as they are used only there. 
- There is an un-used global method defined [here](https://github.com/slaclab/pysmurf/blob/main/python/pysmurf/client/command/cryo_card.py#L25-L30). You should consider removing it, or at least included it as a member of the `CryoCard` class if it is really needed.

ruck314:
Status check on `main` for each of the five points raised here:

1. **Extra unreachable `return` in `do_read`** — Fixed. Resolved in 8323bbc0 (May 2024) which replaced the trailing `return (0)` with an exception, then tightened in d7558c9a (Nov 2025) to `raise RuntimeError(...)`.
[...]

PR #923, opened 2026-05-01
fix(CryoCard): Hide low-level helpers, correct address bit masks, drop dead code (#486)

ruck314:
## Summary
Closes the remaining items in #486:

- Move `cmd_address` / `cmd_data` / `cmd_make` into `CryoCard` as private `@staticmethod` members (`_cmd_address`, `_cmd_data`, `_cmd_make`).
- Tighten the address bit mask to `0x7FF00000` in both helpers. The 32-bit cmd word is: 1 read flag (bit 31), 11 address (bits 20–30), 20 data (bits 0–19). The previous masks (`0x7FFF0000` in `cmd_address`, `0x7FFF00000` in `cmd_make`) were over-broad but produced identical bit patterns for valid inputs, so this is a clarity fix, not a behavioral change.
- Remove unused global `write_csv` (and its lone `import os`).
- Remove unused global `cmd_read` (dead code; never called anywhere in the repo).

Items 1 and 3 of the issue (extra `return` after `raise`, silent `return 0` on read failure) were already addressed in 8323bbc0 and d7558c9a, so they aren't touched here.

## Out of scope (flagged for future cleanup)
- `read_relays` has a `time.sleep(0.1)` after a `return` — pre-existing unreachable code.
- Comment typo `# do we have the right addres` in `read_cycle_count`.

## Test plan
- [x] `flake8 --count python/` — 0 violations.
- [x] Module imports cleanly; the three `_cmd_*` helpers are reachable on the class.
- [x] Verified the new vs. old masks produce identical results for valid inputs (`cmd_address` on any 32-bit value; `cmd_make` for 11-bit addresses).
- [ ] CI green.

### DIFF
--- a/python/pysmurf/client/command/cryo_card.py
+++ b/python/pysmurf/client/command/cryo_card.py
@@ -15,7 +15,6 @@
 # contained in the LICENSE.txt file.
 #-----------------------------------------------------------------------------
 import time
-import os
 
 import numpy as np
 try:
@@ -26,12 +25,6 @@ except ModuleNotFoundError:
 
 from ..base.logger import SmurfLogger
 
-def write_csv(filename, header, line):
-    should_write_header = os.path.exists(filename)
-    with open(filename, 'a+') as f:
-        if not should_write_header:
-            f.write(header+'\n')
-        f.write(line+'\n')
 
 class CryoCard():
     def __init__(self, readpv_in, writepv_in, log=None, server_addr="localhost", server_port=9000):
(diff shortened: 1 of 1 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. Stage 3 replaces the hand-written smurf_command/base_class register layer (and its docstrings) with runtime introspection and a generated shim, so this change would be discarded. Leaving #486 open to be revisited on the new layer.

PR #924, opened 2026-05-01
fix(find_freq): Default to firmware-processed subbands (#482)

### LINKED ISSUE #482: `find_freq` should default to all processed subbands if given no list of subbands
`find_freq` should default to all processed subbands (which may be different depending on fw) if given no list of subbands via the subband keyword.

https://github.com/slaclab/pysmurf/blob/a370b515ab717c982781223da147bea3c8fb3a9c/python/pysmurf/client/tune/smurf_tune.py#L3345

ruck314:
## Summary
- Closes #482. When `find_freq` is called with no `subband` argument, it now sweeps only the subbands the firmware actually processes (derived from `get_number_processed_channels`) instead of all 128 / 512 subbands.
- Adds a `get_processed_subbands(band)` helper next to `get_processed_channels` in `smurf_util.py`, scaling firmware-aware: returns `arange(12, 116)` for 128-subband firmware and `arange(48, 464)` for 512-subband firmware (both equal `0.8125 × n_subbands`).
- In `find_freq`, the default branch (`subband=None`) now intersects the `start_freq`/`stop_freq`-derived subband range with the processed-subband list. Behavior with an explicit `subband=` argument is unchanged.
- Replaces the misleading docstring that referenced a hardcoded `numpy.arange(13,115)` and an inverted "takes precedent" note.

## Test plan
- [x] `flake8 --count python/` passes (matches CI invocation in `.github/workflows/test-or-deploy.yml`).
- [x] Offline math check: `get_processed_subbands` returns expected ranges and lengths for both 128/2048 and 512/8192 firmware variants, with symmetric endpoints.
- [ ] On-hardware: run `S.find_freq(band)` (no `subband` arg) on a 128-subband system; confirm `freq_resp[band]['find_freq']['subband']` excludes 0–11 and 116–127, and that `setup_notches` still consumes `freq_resp` correctly.
- [ ] On-hardware: confirm a narrow user-supplied `start_freq`/`stop_freq` with `subband=None` yields the intersection of the user range and the processed range.
- [ ] Regression: pass an explicit `subband=np.arange(20,40)` and confirm behavior is unchanged.

### DIFF
--- a/python/pysmurf/client/tune/smurf_tune.py
+++ b/python/pysmurf/client/tune/smurf_tune.py
@@ -3452,11 +3452,11 @@ class SmurfTuneMixin(SmurfBase):
             The scan start frequency in MHz (from band center)
         stop_freq : float, optional, default 250
             The scan stop frequency in MHz (from band center)
-        subband : deprecated, use start_freq/stop_freq.
-            numpy.ndarray of int or None, optional, default None
-            An int array for the subbands.  If None, set to all
-            processed subbands =numpy.arange(13,115).
-            Takes precedent over start_freq/stop_freq.
+        subband : numpy.ndarray of int or None, optional, default None
+            An int array of subbands to sweep.  If None, defaults to
+            the intersection of the start_freq/stop_freq range with
+            the firmware's processed subbands (see
+            get_processed_subbands).
         tone_power : int or None, optional, default None
             The drive amplitude.  If None, takes from cfg.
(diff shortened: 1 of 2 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. This changes the find_freq default rather than fixing wrong behaviour, and find_freq is redesigned as a function over a session at stage 6, which is the right place to settle its default. Leaving #482 open.

PR #926, opened 2026-05-01
fix(estimate_phase_delay): Stitch DSP subbands and restore atts on crash (#469)

### LINKED ISSUE #469: DSP full band response unwrap broken (in sw?) with new(er) fw versions
## Describe the bug
Full band response unwrap broken for DSP response in `estimate_phase_delay` function for new fw versions.  Just got this with fw version bdc4981.  Not sure which fw version this started happening for or if it's even a fw issue:

image

Also, need to check if `estimate_phase_delay` function returns atts to nominal when it finishes, either it doesn't or when I just ran it and it unexpectedly crashed I got into a bad state.  I think the latter.

## To Reproduce

Run this command to reproduce the attached plot - `S.estimate_phase_delay(2,freq_min=-10e6,freq_max=10e6)`.

ruck314:
## Summary

Closes #469.

Two bugs in `estimate_phase_delay` reported by @swh76:

1. **Broken DSP unwrap** in the middle plot panel for newer firmware. The first DSP `find_freq` pass concatenated per-subband responses without inter-subband phase stitching, so `np.unwrap(np.angle(resp_dsp_subset))` tripped on the firmware-imposed phase jumps at every subband boundary. Each subband is now rotated so its first sample lines up with the last accumulated sample, mirroring what the second pass already does.

2. **Pre-existing pass-2 stitching bug.** `new_phase = np.angle(resp_dsp_corr[sb][idx[0]])` was a vector, so the multiplication `resp * exp(1j*(last_phase - new_phase))` collapsed every sample of the new subband to `last_phase` instead of rotating it by a single offset. Fixed by indexing the first sample only (`idx[0][0]`) so the rotation is a true rigid stitch. The user only screenshot'd the middle panel in the issue, but the bottom 'Residuals' panel was silently affected too.

3. **Attenuators not restored on crash.** `uc_att`/`dc_att` were saved at the top and restored only at the very bottom, so any exception in `find_freq` / `full_band_resp` / polyfit / plotting left the band programmed with the function's att values instead of the caller's. Body is now wrapped in `try` / `finally`; the `finally` holds only the two `set_att_*` calls, so the success path is byte-identical. `plt.show()` was moved inside `try` so the `finally` remains the last action.

## Changes

`python/pysmurf/client/util/smurf_util.py` — `estimate_phase_delay()` only. No new helpers, no API surface change.

## Out of scope

- Issue #513 / PR #913 (resonator-free `phaseDelaySubbands` config field). Independent feature, separate branch already in flight.
- Restoring `band_delay_us` to its pre-call value — the function's documented contract is to **set** it to the estimated optimum on completion.
- Restoring `band_off` / `flux_ramp_off` — intentional setup, no documented restore contract, not flagged in #469.
- Cleaning up the dead-code global polyfit at line 468 whose result is overwritten by `np.mean(est_delay)` on the next line. Pre-existing cruft, unrelated.

## Test plan

- [x] `flake8 --count python/` → 0 (matches CI invocation in `.github/workflows/test-or-deploy.yml`)
- [x] `python3 -m py_compile python/pysmurf/client/util/smurf_util.py` — clean
- [x] Offline math sanity for the stitching fix: synthesized two complex subbands with a 7.3 rad phase offset between them; confirmed (a) old code flattens every stitched sample to `last_phase`, (b) new code produces a monotonically unwrappable concatenated phase with the boundary jump driven to 0 and the polyfit slope recovered to the input slope.
- [ ] On-hardware (reporter's exact reproducer): `S.estimate_phase_delay(2, freq_min=-10e6, freq_max=10e6)` — confirm middle "DSP Delay" panel shows a smooth ~linear unwrapped phase across all subbands, bottom "Residuals" panel shows reasonable post-correction residuals, and originally programmed UC/DC att values are restored.
- [ ] On-hardware crash path: induce an exception inside the function (e.g., interrupt during `find_freq`); confirm UC/DC atts return to their pre-call values.

### DIFF
--- a/python/pysmurf/client/util/smurf_util.py
+++ b/python/pysmurf/client/util/smurf_util.py
@@ -365,238 +365,249 @@ class SmurfUtilMixin(SmurfBase):
         self.set_att_uc(band, uc_att, write_log=True)
         self.set_att_dc(band, dc_att, write_log=True)
 
-        # only loop over dsp subbands in requested frequency range (to
-        # save time)
-        n_subbands = self.get_number_sub_bands(band)
-        digitizer_frequency_mhz = self.get_digitizer_frequency_mhz(band)
-        subband_half_width_mhz = digitizer_frequency_mhz/\
-            n_subbands
-        subbands,subband_centers=self.get_subband_centers(band)
-        subband_freq_min=-subband_half_width_mhz/2.
-        subband_freq_max=subband_half_width_mhz/2.
-        dsp_subbands=[]
-        for sb,sbc in zip(subbands,subband_centers):
-            # ignore unprocessed sub-bands
-            if sb not in subbands:
-                continue
-            lower_sb_freq=sbc+subband_freq_min
-            upper_sb_freq=sbc+subband_freq_max
-            if lower_sb_freq>=(freq_min/1.e6-subband_half_width_mhz) and \
-                    upper_sb_freq<=(freq_max/1.e6+subband_half_width_mhz):
-                dsp_subbands.append(sb)
+        try:
+            # only loop over dsp subbands in requested frequency range (to
+            # save time)
+            n_subbands = self.get_number_sub_bands(band)
+            digitizer_frequency_mhz = self.get_digitizer_frequency_mhz(band)
+            subband_half_width_mhz = digitizer_frequency_mhz/\
+                n_subbands
+            subbands,subband_centers=self.get_subband_centers(band)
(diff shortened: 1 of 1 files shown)


ruck314:
Closing as mostly landed. Main already stitches the corrected-response subbands in estimate_phase_delay, so only the uncorrected-response stitching and the attenuator-restoring try/finally remain, and those are better redone as a small fresh change against current main than by rebasing this one. Leaving #469 open for that.

PR #891, opened 2026-04-30
docs(set_feedback_limit_khz): Fix and expand docstring (#762)

ruck314:
## Summary

Closes #762.

The current docstring for `SmurfUtilMixin.set_feedback_limit_khz` is wrong on two of its three lines:

- `band : int` is described as "The band that is to be turned off" — clearly copy-pasted from `band_off`/`channel_off`.
- `feedback_limit_khz` is described as a "feedback rate", but it's a tracking-loop frequency deviation limit, not a rate.
- The summary doesn't mention the silent clamping to half the subband bandwidth, the 16-bit decimal encoding, or how this relates to the underlying `feedbackLimit` register.

This PR rewrites the docstring to:
- Correctly describe `band` as the 500 MHz band being configured.
- Correctly describe `feedback_limit_khz` as a tracking-loop frequency limit, in kHz.
- Explain the constraint enforced by the feedback loop (`abs(feedbackFreq) < feedbackLimit`).
- Document the silent clamping to half the subband bandwidth and the 16-bit decimal encoding written to the `feedbackLimit` register via `set_feedback_limit`.
- Cross-reference `set_feedback_limit` / `get_feedback_limit`.

No code logic was changed — docstring-only.

## Test plan

- [x] `flake8 --count python/pysmurf/client/util/smurf_util.py` passes (matches the CI invocation in `.github/workflows/test-or-deploy.yml`).
- [ ] CI: flake8, docs build, server tests, client tests.

PR #928, opened 2026-05-01
docs(set_ramp_rate): Clarify timing-system-only use (#465)

### LINKED ISSUE #465: `set_ramp_rate` is only for use with a timing system.  Need to clarify this.
## Describe the bug
The `smurf_command.set_ramp_rate` routine probably needs to be renamed and its docstring needs to be updated to make it clear that this command is only to be used with a timing system.

ruck314:
## Summary
- Closes #465.
- Rewrite the docstrings of `set_ramp_rate` and `get_ramp_rate` in `python/pysmurf/client/command/smurf_command.py` to make it explicit that both routines write/read the EVR timing trigger register (`AMCc.…AmcCarrierTiming.EvrV2CoreTriggers.EvrV2ChannelReg[0].RateSel`) and only affect the actual flux ramp rate when the carrier is configured to use the timing system trigger (`get_ramp_start_mode() == 0x1`). In internal (`0x0`) or external (`0x2`) trigger modes the reset rate is set by `flux_ramp_setup` via `set_ramp_max_cnt`, so these routines are no-ops for the actual ramp.
- Cross-reference `get_ramp_start_mode`, `flux_ramp_setup`, `set_ramp_max_cnt`, `flux_ramp_rate_to_PV`, and `flux_ramp_PV_to_rate` so a confused user has an obvious next step.
- The numpydoc-style sections (`Args` / `Returns` / `See Also`) match the convention used elsewhere in `smurf_command.py`.

## Why docs-only, not rename
Issue #465's "probably needs to be renamed" is a soft suggestion; the firm ask is "docstring needs to be updated to make it clear." `set_ramp_rate` is part of the public `SmurfCommandMixin` API and is referenced from `smurf_tune.flux_ramp_setup` and downstream user code, so renaming would be breaking. The single internal caller at `smurf_tune.py:3265` already gates `set_ramp_rate` behind `if self.get_ramp_start_mode() == 1:`, so no behavioral change is needed — only the documentation was misleading. Repo convention for similar issues is a `docs(...)` PR (see PR #910, #891).

## Test plan
- [x] `flake8 --count python/` passes (matches the CI invocation in `.github/workflows/test-or-deploy.yml`); `flake8-rst-docstrings` parses the new RST cleanly.
- [x] AST extraction of the new docstrings renders the warning block, `Args`/`Returns`, and `See Also` cross-references correctly.
- [x] No source other than the two docstrings is touched; no rename, no runtime guard, no behavioral change.

### DIFF
--- a/python/pysmurf/client/command/smurf_command.py
+++ b/python/pysmurf/client/command/smurf_command.py
@@ -3868,11 +3868,43 @@ class SmurfCommandMixin(SmurfBase):
     _trigger_rate_sel_reg = ".EvrV2ChannelReg[0].RateSel"
 
     def set_ramp_rate(self, val, **kwargs):
-        """
-        flux ramp sawtooth reset rate in kHz
+        r"""Set the EVR timing-trigger flux ramp reset rate, in kHz.
 
-        If using timing system, the allowed rates are: 1, 2, 3, 4, 5,
-        6, 8, 10, 12, 15kHz (hardcoded by timing)
+        .. warning::
+           This routine **only takes effect when the carrier is
+           configured to use the timing system trigger**, i.e. when
+           :func:`get_ramp_start_mode` returns ``0x1``.  In internal
+           (``0x0``) or external (``0x2``) trigger modes the flux ramp
+           reset rate is set by :func:`flux_ramp_setup` via
+           :func:`set_ramp_max_cnt`, and calling this routine has no
+           effect on the actual ramp rate.
+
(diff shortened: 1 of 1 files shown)


swh76:
Closing: these docstring updates for set_ramp_rate and get_ramp_rate are already on main (merged via an earlier doc branch). Main's versions are equivalent in content and also add kwargs/return type documentation that this branch lacks. No unique changes remain to merge.

PR #940, opened 2026-05-01
fix(relock): Rename eta_mag to eta_mag_scaled and clarify etaMag docs (#416)

### LINKED ISSUE #416: eta_mag_scaled is not eta_mag in relock
## Describe the bug
The variable defined as eta_mag here:

https://github.com/slaclab/pysmurf/blob/3d3271116ab8f8d7384e0d5179b9e9d029408586/python/pysmurf/client/tune/smurf_tune.py#L1936

is *not* eta_mag.  It is eta_mag_scaled.  We also need to find a way to make it more obvious that the `set_eta_mag_array` function actually sets the array of etaMagScaled values, not the array of etaMag values.

ruck314:
## Summary
- Fixes [#416](https://github.com/slaclab/pysmurf/issues/416). In `relock()`, a local variable named `eta_mag` was populated from `freq_resp[...]['eta_scaled']` and passed to `set_eta_mag_array` — values were correct, name was wrong.
- Renamed the local variable to `eta_mag_scaled` in `relock()` (three sites in `smurf_tune.py`).
- Expanded the empty docstrings on `set_eta_mag_array` / `get_eta_mag_array` to make it explicit that the firmware `etaMag` register actually holds *scaled* values (`eta_mag / subband_half_width`), so the misleading function name is at least documented.
- No behavioral change; values written to the register are unchanged.

## Out of scope
A second instance of the same naming bug exists at `smurf_tune.py:4855` (`band_outputs[band]['eta_mag'] = list(self.get_eta_mag_array(band))`). Renaming that dict key would change the schema of status-dump config files and break any downstream consumer reading the `'eta_mag'` key, so it's left for a follow-up.

## Test plan
- [x] `flake8 --count python/` passes (0 errors) — same command CI runs.
- [ ] CI flake8 / docs / server / client jobs pass on the PR.
- [ ] Spot-check on hardware that `relock()` still tunes channels as expected (pure rename, so behavior is unchanged).

### DIFF
--- a/python/pysmurf/client/command/smurf_command.py
+++ b/python/pysmurf/client/command/smurf_command.py
@@ -2118,6 +2118,11 @@ class SmurfCommandMixin(SmurfBase):
 
     def set_eta_mag_array(self, band, val, **kwargs):
         """
+        Sets the band-array etaMag firmware register. Despite the name,
+        this register holds *scaled* etaMag values, i.e.
+        eta_mag / subband_half_width, where
+        subband_half_width = digitizer_frequency_mhz / n_subbands.
+        Pass eta_mag_scaled values, not unscaled eta_mag.
         """
         self._caput(
             self._cryo_root(band) + self._eta_mag_array_reg,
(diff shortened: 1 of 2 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. Stage 3 replaces the hand-written smurf_command/base_class register layer (and its docstrings) with runtime introspection and a generated shim, so this change would be discarded. Leaving #416 open to be revisited on the new layer.

PR #912, opened 2026-04-30
fix(noise_vs_amplitude): Add analyze_noise_vs_amplitude and remove deprecated noise_vs_tone (#521)

ruck314:
## Summary

Closes #521.

`noise_vs_amplitude` delegated to the generic `noise_vs()` dispatcher, which **unconditionally** called `analyze_noise_vs_bias` regardless of the swept variable. Because no `analyze_noise_vs_amplitude` method existed, the amplitude analyze path silently ran the TES-bias analyzer on amplitude-sweep data.

This PR adds the missing analyzer, branches the dispatcher per-`var`, and removes the long-deprecated `noise_vs_tone` / `analyze_noise_vs_tone` (zero callers since 2020). The `noise_vs_bias` path is byte-identical — the bias-branch call is unchanged.

## Changes

- **`python/pysmurf/client/debug/smurf_noise.py`**
  - Add `analyze_noise_vs_amplitude(amplitude, datafile, ...)`. Ports the PSD load / smooth / fit / plot logic from the (removed) `analyze_noise_vs_tone`. Drops the meaningless `bias_group` plumbing, uses the `Tone amplitude [unit-less]` axis label, writes `noise_vs_amplitude_band{band}_ch{ch:03}.png` plots, fixes the latent malformed f-string `{freq_max:.2f Hz}` → `{freq_max:.2f} Hz`, and uses `self.get_sample_frequency()` (matches `analyze_noise_vs_bias`) instead of `self._fs`.
  - Branch `noise_vs()`'s analyze block per `var`: bias path byte-identical; amplitude path calls the new analyzer.
  - Drop the dead `kwargs['bias_group']=-1` hack that only existed to make the unconditional bias-analyzer call work for the amplitude path.
  - Remove deprecated `noise_vs_tone` and `analyze_noise_vs_tone`.
  - Drop now-orphaned `import time`.
- **`python/pysmurf/client/base/smurf_config_properties.py`**
  - Repoint the `pA_per_phi0` Sphinx `:func:` xref from `analyze_noise_vs_tone` to `analyze_noise_vs_amplitude` (the semantic replacement).

Net diff: +251 / -366 lines.

## Test plan

- [x] `python -m py_compile` on both modified files — clean
- [x] `flake8 --count python/` (matches `.github/workflows/test-or-deploy.yml` invocation) — 0 issues
[...]

PR #943, opened 2026-05-01
fix(noise_vs_tone): Default fraction_full_scale to None (#355)

### LINKED ISSUE #355: noise_vs_tone (and other functions?) using odd default args
`noise_vs_tone` and probably other functions are using some very specific and dumb defaults ; e.g. fraction_full_scale=.72 :

https://github.com/slaclab/pysmurf/blob/6a35876e369d227e907b7df3f95655473749060f/python/pysmurf/client/debug/smurf_noise.py#L330

it should instead have the default fraction_full_scale=None, which then pulls whatever's in the config or stored at runtime from the last `tracking_setup` call.  Whoever closes this issue should scan over pysmurf functions to look for functions with similarly bad defaults.

ruck314:
## Summary

Fixes #355.

`noise_vs_tone` shipped with the magic-number default `fraction_full_scale=.72`, which silently overrode the cfg value `tune_band:fraction_full_scale` whenever a caller did not explicitly pass the argument — defeating the fallback that `tracking_setup` already implements.

- Change `noise_vs_tone` signature default from `fraction_full_scale=.72` to `fraction_full_scale=None` (`python/pysmurf/client/debug/smurf_noise.py:366`).
- Update the docstring to note the cfg fallback.

No body changes were needed — `noise_vs_tone` already passes `fraction_full_scale` straight into `tracking_setup`, and `tracking_setup` (`python/pysmurf/client/tune/smurf_tune.py:2643-2644`) does:

```python
if fraction_full_scale is None:
    fraction_full_scale = self._fraction_full_scale
```

`self._fraction_full_scale` is loaded from `tune_band:fraction_full_scale` in `smurf_config_properties.py:330`.

## Out of scope

The issue also asks to scan for other functions with similarly bad defaults. Findings:

- `optimize_lms_delay` (`smurf_tune.py:4328`) already uses `fraction_full_scale=None` — correct.
- `git grep "fraction_full_scale=\.72"` returns zero library hits after this fix (only `scratch/eyy/debug/nonlinear_lms.py` remains, which is debug code, not the library).
- `n_phi0=4`, `meas_time=30`, `bias_high=1.5`, etc. were debated in the 2020 issue thread (eyyoung24 vs swh76) without resolution. swh76 leaned toward keeping `meas_time` hardcoded but moving `n_phi0` to cfg. `n_phi0` is not currently a cfg property anywhere in `smurf_config_properties.py`; promoting it would require a new cfg key plus updates to `tracking_setup`, `optimize_lms_delay`, and several `noise_vs_*` callers — a separate refactor.

## Test plan

- [x] `flake8 --count python/` (CI lint command) → 0 errors
- [x] `git grep "fraction_full_scale=\.72\|fraction_full_scale=0\.72" -- python/pysmurf/` → no library hits
- [x] Confirmed `noise_vs_tone` body still passes `fraction_full_scale` straight to `tracking_setup` (line 432, unchanged)
- [x] Confirmed `tracking_setup`'s `None` → `self._fraction_full_scale` fallback path is intact
- [ ] CI: docker server/client tests on PR

### DIFF
--- a/python/pysmurf/client/debug/smurf_noise.py
+++ b/python/pysmurf/client/debug/smurf_noise.py
@@ -363,7 +363,7 @@ class SmurfNoiseMixin(SmurfBase):
 
     def noise_vs_tone(self, band, tones=None, meas_time=30,
                       analyze=False, bias_group=None, lms_freq_hz=None,
-                      fraction_full_scale=.72, meas_flux_ramp_amp=False,
+                      fraction_full_scale=None, meas_flux_ramp_amp=False,
                       n_phi0=4, make_timestream_plot=True,
                       new_master_assignment=True, from_old_tune=False,
                       old_tune=None):
@@ -387,8 +387,10 @@ class SmurfNoiseMixin(SmurfBase):
         lms_freq_hz : float or None, optional, default None
             The tracking frequency in Hz. If None, measures the
             tracking frequency.
-        fraction_full_scale : float, optional, default 0.72
-            The amplitude of the flux ramp.
+        fraction_full_scale : float or None, optional, default None
+            The amplitude of the flux ramp. If None, defaults to the
+            cfg value ``tune_band:fraction_full_scale`` (via
+            ``tracking_setup``).
         meas_flux_ramp_amp : bool or None, optional, default False
             Whether to measure the flux ramp amplitude.
         n_phi0 : float, optional, default 4.0


ruck314:
Closing as superseded by #912, which removes noise_vs_tone and analyze_noise_vs_tone entirely (no callers since 2020), so this default fix has nothing left to apply to. Leaving #355 open.

PR #944, opened 2026-05-01
feat: Add **kwargs to outward-facing functions for forward-compat (#385)

### LINKED ISSUE #385: add **kwargs to functions that are outward facing
## Describe the problem
For outward facing functions that we want to eventually freeze the interface, we want to be able to add more optional variables. But to preserve backwards functionality, the old functions need to accept new arguments.

Add `**kwargs` to the argument list in a function. This means if we add a argument in a new branch, the old function will just accept the new argument without crashing (but it will also not do anything).

ruck314:
## Summary
- Closes #385
- Adds `**kwargs` to 112 high-level user-entry points across the four user-facing mixins (`smurf_tune.py`, `smurf_util.py`, `smurf_iv.py`, `smurf_noise.py`)
- Future feature branches can introduce new optional arguments to these functions without breaking older callers — exactly the design described in the issue
- Signature-only change. Zero behavior change. Each function silently absorbs unknown kwargs (matches the existing pattern in `check_lock`, `noise_vs`, `toggle_feedback`, `band_off`)

## Coverage by file (public methods, AST-verified)
| File | Before (with kwargs) | After (with kwargs) |
|---|---|---|
| `smurf_tune.py` | 2 / 63 | 46 / 63 |
| `smurf_util.py` | 13 / 95 | 54 / 95 |
| `smurf_iv.py` | 1 / 7 | 7 / 7 |
| `smurf_noise.py` | 1 / 24 | 22 / 24 |

Combined with `smurf_command.py` (already 92% covered since 2020) and `smurf_atca_monitor.py` (100%), all major user-facing entry points now accept `**kwargs`.

## Out of scope (deliberately)
- `smurf_command.py` — already 92% covered; the remaining 35 are mostly deprecation shims being phased out in adjacent PRs
- Thin PV getters with ≤1 arg
- Pure math/converter helpers (`freq_to_band`, `channel_to_freq`, etc.) — they don't accumulate parameters over time, so `**kwargs` adds no value and only masks typos
- `smurf_control.py` low-level utilities (`make_dir`, `get_timestamp`, `add_output`, `write_output`)

## Test plan
- [x] `flake8 --count python/` passes (0 errors, full CI invocation)
- [x] `python -m py_compile` on all 4 files succeeds
- [x] AST coverage check: all 112 target methods now have `**kwargs`
- [ ] Hardware-dependent tests in `python/pysmurf/client/test/` require a live cryostat — not run here. Risk is zero because the change is signature-only.

### DIFF
--- a/python/pysmurf/client/debug/smurf_iv.py
+++ b/python/pysmurf/client/debug/smurf_iv.py
@@ -34,7 +34,7 @@ class SmurfIVMixin(SmurfBase):
                channels=None, band=None, high_current_mode=True,
                overbias_voltage=8., grid_on=True,
                phase_excursion_min=3., bias_line_resistance=None,
-               do_analysis=True):
+               do_analysis=True, **kwargs):
         """Takes a slow IV
 
         Steps the TES bias down slowly. Starts at bias_high to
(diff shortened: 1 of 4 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. Stage 6 dissolves the SmurfControl mixins into functions over a session, so a signature sweep across the mixins would be discarded. Leaving #385 open.

PR #946, opened 2026-05-01
feat(offline): Read register values from rogue state dump (#364)

### LINKED ISSUE #364: Add support for an offline mode that reads back a state dump
## Describe the problem

Currently, offline mode only supports a handful of gets using hard-codes like:

https://github.com/slaclab/pysmurf/blob/37a028b6f656eb0c00d9e472470bdd2476a29819/python/pysmurf/client/command/smurf_command.py#L1647

but we could have pysmurf instead read a state dump and treat it like it's a rogue server, which would be more flexible, and potentially enable using more gets & help facilitate offline development.

ruck314:
## Summary

Closes #364 (open since 2020-03-26). Pysmurf offline mode previously returned `None` from `_caget` and relied on a handful of per-method hard-coded constants (two are literally commented `# FIX ME - STUPPID HARDCODE`). The issue asks for pysmurf to read a rogue `SaveState` YAML dump and *"treat it like it's a rogue server"* so offline development can exercise more gets.

This PR plumbs a captured state dump into offline mode end-to-end:

- `tools.load_state_yaml(filename)` — loads a rogue SaveState dump, transparently handling `.yml` and `.yml.gz` (rogue's `saveYaml(..., autoCompress=True)` default).
- `tools.state_lookup(state, pvname, default=None)` — walks a dot-separated rogue PV path through the nested dict; handles indexed segments like `Base[0]` as either a literal key or a list/dict index, since rogue versions vary.
- `SmurfBase.__init__(..., state_file=None, ...)` — new kwarg loads the dump on construction into `self._offline_state`.
- `SmurfControl.__init__(..., state_file=None, ...)` — passes through.
- `SmurfCommandMixin.load_state(filename)` — runtime loader symmetric with the existing `save_state`.
- `_caget` offline branch consults the dump first; falls back to the existing `None` return when the key is missing or no dump is loaded.
- 9 hard-coded `if self.offline:` sites guarded with `and self._offline_state is None` so a loaded dump preempts the constant:

  | File | Method | Hardcode |
  |---|---|---|
  | smurf_command.py | `get_number_sub_bands`        | 128 |
  | smurf_command.py | `get_number_channels`         | 512 |
  | smurf_command.py | `get_band_center_mhz`         | 4250 + band*500 |
  | smurf_command.py | `get_channel_frequency_mhz`   | 2.4 |
  | smurf_command.py | `get_digitizer_frequency_mhz` | 614.4 |
  | smurf_command.py | `get_flux_ramp_freq`          | 4.0 |
  | smurf_command.py | `get_filter_a`                | 5-elt ndarray |
  | smurf_command.py | `get_filter_b`                | 5-elt ndarray |
  | smurf_command.py | `get_downsample_factor`       | 20 |

Backward compatible: with no `state_file`, behavior is byte-identical to today.

## Usage

```python
S = SmurfControl(offline=True,
                 cfg_file='cfg/experiment.cfg',
                 state_file='/data/state.dat.yml.gz')
S.get_digitizer_frequency_mhz(0)   # recorded value, not 614.4
S.load_state('/data/newer.yml.gz')  # swap at runtime
```

## Tradeoff (documented on `state_file`)

A loaded dump that lacks a key yields `None` rather than the legacy hardcode. Intentional — the user opted into the dump, and a missing key is a real signal. If this proves brittle in practice, a follow-up can add a per-method fallback.

## Test plan

- [x] New `python/pysmurf/client/test/test_offline_state.py` — 9 tests: `state_lookup` walk, missing-key default, gzip/plain loading, `_caget` returns `None` without dump, reads from dump with one, `load_state` swaps at runtime, hardcodes used without a dump, dump overrides hardcodes, missing-key returns `None` (documented). All pass; no pyrogue / hardware required.
- [x] `flake8 --count python/` → 0 (matches CI invocation).
- [ ] CI flake8 / docs / server / client jobs pass on the PR.
- [ ] Manual hardware regression — operator captures a real `state.*.yml.gz` via `S.save_state(...)` on a live system, restarts pysmurf with `offline=True, state_file=<that file>`, and confirms `_caget`-backed methods return recorded values. (Out of scope to capture here.)

## Out of scope

- Removing `tools.yaml_parse` (dead code: splits on `:` while pysmurf pvnames use `.`; zero call sites). Left untouched to avoid breaking any external script that may still import it.
- Refactoring the 9 hard-coded methods to fully delegate to `_caget`. The single-line guard preserves the legacy hardcode when no dump is loaded; a follow-up PR can fold the constants into `_caget`-backed defaults if desired.

### DIFF
--- a/python/pysmurf/client/base/base_class.py
+++ b/python/pysmurf/client/base/base_class.py
@@ -48,6 +48,12 @@ class SmurfBase:
     offline : bool, optional, default False
         Whether to run in offline mode (no rogue) or not. This
         will break many things. Default is False.
+    state_file : str or None, optional, default None
+        Path to a rogue ``SaveState`` YAML dump (``.yml`` or
+        ``.yml.gz``) to back ``_caget`` calls in offline mode.  Only
+        consulted when ``offline`` is True.  When loaded, recorded
+        register values take precedence over the per-method offline
+        hard-coded fallbacks; missing keys yield ``None``.
     pub_root : str or None, optional, default None
         Root of environment vars to set publisher options. If
         None, the default root will be `SMURFPUB_`.
@@ -77,7 +83,8 @@ class SmurfBase:
     """
 
     def __init__(self, log=None, server_addr="localhost", server_port=9000, atca_port=9100,
-                 atca_monitor=False, offline=False, pub_root=None, script_id=None, **kwargs):
+                 atca_monitor=False, offline=False, state_file=None, pub_root=None,
+                 script_id=None, **kwargs):
         """
         """
 
@@ -124,6 +131,15 @@ class SmurfBase:
         self.offline = offline
         if self.offline is True:
             self.log('Offline mode')
+
+        # Optional rogue SaveState YAML dump used to back _caget calls
+        # in offline mode.  Populated either here from the state_file
+        # constructor argument, or later via SmurfCommandMixin.load_state().
+        self._offline_state = None
+        if offline and state_file is not None:
+            from pysmurf.client.util import tools
+            self._offline_state = tools.load_state_yaml(state_file)
+            self.log(f'Offline state loaded from {state_file}')
 
         # Setting paths for easier commands - Is there a better way to
         # do this than just hardcoding paths? This needs to be cleaned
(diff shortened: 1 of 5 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. Stage 4 replaces the client config layer with a single inherited schema, and the resolved config is published by the server (plus sidecar), which covers this use case. Leaving #364 open.

PR #947, opened 2026-05-01
feat(SmurfControl): Dump loaded pysmurf cfg into output_dir on startup (#286)

### LINKED ISSUE #286: dump config on startup
During SO testing and operation we'll probably want to have a record of the config state of pysmurf for a given set of data. Right now basically a single config file is being maintained, but it's hard to know if it pertains to data that was taken in the past or if it has been changed. We'll probably want to do something like dump the config file each time a new output directory is created, so we can have a first order approximation of what the pysmurf state was for a given set of data. This won't tell us about changes that were made in pysmurf between setup and data-taking, but I think it is a good step. 

Is this prohibitively time-consuming? If not I think we should add an init argument that will tell pysmurf to dump the config file after setup is done. I imagine we'll want this to be True by default if the data files aren't too large, but we can set that in an SO-specific config file somewhere if you don't want that for everyone.

ruck314:
## Summary
- Adds `dump_config=True` kwarg to `SmurfControl.initialize()` so the loaded `SmurfConfig` is written to `<output_dir>/<start_time>_startup_pysmurf_cfg.cfg` immediately after the run output directory is created.
- Applies in both the `smurf_cmd_mode` branch and the standard branch of `initialize()`.
- Uses the existing `SmurfConfig.write()` primitive — no new dependencies, no changes to other call sites.
- Default `True` matches @jlashner's request in #286; opt-out via `dump_config=False` (or by passing through `SmurfControl(..., dump_config=False)` since `__init__` forwards `**kwargs` to `initialize()`).

Closes #286.

## Test plan
- [x] `python -m py_compile python/pysmurf/client/base/smurf_control.py`
- [x] `SmurfConfig` write→read round-trip on `cfg_files/template/template.cfg` (6.7 KB JSON, dict-equal after reload)
- [ ] Smoke test on a real SMuRF carrier: instantiate `SmurfControl(cfg_file=..., setup=False)` and confirm `<output_dir>/<start_time>_startup_pysmurf_cfg.cfg` exists and parses back via `SmurfConfig`
- [ ] Confirm `dump_config=False` suppresses the dump
- [ ] Spot-check that the startup `.cfg` is created even when `setup=True` raises (it is written before the optional `setup()` call)

## Notes
- Out of scope (per @jlashner's clarification on #286): the full rogue tree dump from `dump_state` — that remains a separate concern.
- The dump filename uses `self.start_time` so it correlates with the run name rather than re-timing via `get_timestamp()`.

### DIFF
--- a/python/pysmurf/client/base/smurf_control.py
+++ b/python/pysmurf/client/base/smurf_control.py
@@ -136,7 +136,7 @@ class SmurfControl(SmurfCommandMixin,
                    make_logfile=True, setup=False,
                    smurf_cmd_mode=False, no_dir=False, publish=False,
                    payload_size=2048, data_path_id=None,
-                   **kwargs):
+                   dump_config=True, **kwargs):
         """Initializes SMuRF system.
 
         Longer description of initialize routine here.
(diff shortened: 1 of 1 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. Stage 4 replaces the client config layer with a single inherited schema, and the resolved config is published by the server (plus sidecar), which covers this use case. Leaving #286 open.

Issue #626, opened 2021-04-20
Efficiency of .dat files


jlashner:
## Describe the bug
For some reason .dat files that are written on SO systems are very space inefficient. For example: Max streamed 4 channels over the weekend, which created a  320GB .dat file, while the g3 data was just 16 GB. This is an ongoing problem at pretty much all test institutions which causes disk space to quickly fill up unreasonably fast.

Matthew also created a script that analyzes the contents of the files, which outputs the following for a standard file: 

```
Reading /mnt/so1/users/jlashner/dat_files/uchicago_missing_data.dat...
  On frame 1, payload size 8320

Analysis:

  Number of frames by "channel":
    ChannelId.Processed           :    238757
    ChannelId.Metadata            :      1193

  Spatial analysis:
    metadata    :    186108     0.0%
    header      :  30560896     1.5%
    stream_data :  93592744     4.7%
    padding     : 1862304600    93.7%

  Summary of FrameCounter steps:
        20 :     238722  99.99%
        26 :          1   0.00%
        25 :          5   0.00%
        29 :          4   0.00%
        27 :          2   0.00%
        24 :          6   0.00%
        31 :          2   0.00%
        30 :          2   0.00%
        32 :          1   0.00%
        23 :          3   0.00%
        33 :          2   0.00%
        35 :          1   0.00%
        22 :          2   0.00%
        37 :          1   0.00%
        38 :          1   0.00%
        21 :          1   0.00%
    other  :          0   0.00%

  Data frame count is    238757 frames in 1193.852 seconds
  Internal frequency     3999.967 Hz
  Readout frequency      199.998 Hz
  Channel counts encountered: {98}
```

Note that 93% of the file is padding. Comparing G3 files with .dat files, it seems like when taking data with `stream_data_on` and `take_stream_data`, the PayloadSize is not being set to match the channel count, so each frame is large enough to hold 4096 channels even though there are only a few real data channels. I think this should be a pretty easy fix to the stream-data functions to just add a line setting the PayloadSize correctly.

jesusvasquez333:
Note that you can set the `PayloadSize ` register under the `ChannelMapper` to `0` in order to make the output frame to contain only the number of channels defined in the `Mask` list (i.e. no padding), as described here:
https://github.com/slaclab/pysmurf/blob/main/README.SmurfProcessor.md#channelmapper

jlashner:
That's easy, is there any reason we'd wouldn't want to just keep the payload size at 0?

[...]

PR #948, opened 2026-05-01
feat(setup): Make payload_size configurable from pysmurf cfg file (#299)

### LINKED ISSUE #299: payload size default in setup
Currently the default payload size is 4096 (when rogue comes up). We should have it set as a smaller number in `S.setup()` and have the value be something taken from the config file.

ruck314:
## Issue
This PR resolves #299.

## Summary
Adds an optional top-level `payload_size` key to the pysmurf configuration file and wires it through `SmurfConfigPropertiesMixin` so `S.setup()` (and `S.initialize()`) use it as the default payload size — fulfilling the second half of #299 that was never addressed (PR #632 only touched `stream_data_on()` for #626).

## Behavior
Precedence at `S.setup()` time:
1. Explicit `payload_size=` kwarg passed by the caller
2. `payload_size` value from the pysmurf cfg file
3. Built-in fallback of `2048` (unchanged behavior when the cfg key is absent)

A value of `0` lets rogue auto-size the payload from the channel mask, matching the convention from #632. Schema validates the value as an int in `[0, 4096]`.

## Files changed
- `python/pysmurf/client/base/smurf_config.py` — `Optional('payload_size', default=None)` schema entry
- `python/pysmurf/client/base/smurf_config_properties.py` — `_payload_size` init, `copy_config_to_properties()` wiring, and getter/setter property
- `python/pysmurf/client/base/smurf_control.py` — `setup()`/`initialize()` defaults changed from `2048` → `None`; precedence resolution added; docstrings updated
- `cfg_files/template/template.cfg` — documented the new optional key

## Verification (offline, before push)
- **Schema** — 6/6 cases pass: template default (2048), explicit override (512), auto-size (0), omitted (→ None), value > 4096 rejected, negative rejected
- **Mixin propagation** — config value flows through `SmurfConfig.get('payload_size')` → `S._payload_size` → `S.payload_size` getter
- **Precedence logic** — 6/6 cases pass, including edge cases (explicit `0` kwarg honored over a non-zero cfg value)
- **Lint** — `flake8 --count python/` (the exact CI command) → 0 errors

## Backward compatibility
- Existing site `.cfg` files (slac, stanford, ucsd, nist, pton, etc.) are not modified.
- The new key is `Optional` with `default=None`; sites that don't set it keep getting `payload_size=2048` from `S.setup()`, identical to today.
- All existing call sites of `S.setup(payload_size=N)` and `S.initialize(payload_size=N)` continue to work unchanged.

- [ ] Yes
- [x] No

### DIFF
--- a/cfg_files/template/template.cfg
+++ b/cfg_files/template/template.cfg
@@ -291,5 +291,10 @@
 
 "default_data_dir": "/data/smurf_data",
 "tune_dir" : "/data/smurf_data/tune",
-"status_dir" : "/data/smurf_data/status"
+"status_dir" : "/data/smurf_data/status",
+
+# Default payload size used by S.setup() (and S.initialize()).  If
+# omitted, S.setup() falls back to 2048.  Set to 0 to let rogue
+# auto-size the payload from the channel mask.  See issue #299.
+"payload_size" : 2048
 }
--- a/python/pysmurf/client/base/smurf_config.py
+++ b/python/pysmurf/client/base/smurf_config.py
@@ -761,6 +761,17 @@ class SmurfConfig:
                                           lambda f: 0 <= f <= 99)
         #### Done specifying thermal schema
 
+        #### Start payload size schema
+        # Default payload size used by S.setup().  If None or absent,
+        # S.setup() falls back to its built-in default (2048).  A value
+        # of 0 lets rogue auto-size the payload from the channel mask.
+        # See issue #299.
+        schema_dict[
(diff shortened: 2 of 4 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. Stage 4 replaces smurf_config.py / smurf_config_properties.py / template.cfg and the cfg_files layout with a single inherited config schema, so this belongs in the new schema rather than in code slated for deletion. Leaving #299 open.

PR #949, opened 2026-05-01
feat(take_noise_psd): Save fitted wl and n columns to .txt output (#186)

### LINKED ISSUE #186: Add option to output fitted white noise level from take_noise_psd
Currently take_noise_psd returns a .dat and mask file to the filepath/outputs directory. If the save_data argument is passed as true an additional .txt file is returned into the plots directory that has a list of frequency, median noise between low_freq + high_freq, and the fitted f_knee. Since the plots already fit for the white noise level, the knee, and the index it makes sense to output those fitted parameters in this optional saved output text file (since this fitted value can be different from the median returned if there is significant 1/f in the data) so that you don't have to go back and rerun the fitting functions on the raw data output to get these data in post processing.

ruck314:
## Summary

Closes #186.

`take_noise_psd` already fits a white-noise + 1/f model (`wl`, `n`, `f_knee`) per channel and renders those values on the per-channel plot, but when `save_data=True` the resulting `.txt` only contained `res_freq`, `noise_floor`, and `f_knee`. Post-processing therefore had to reload the raw timestream and re-run the fit just to recover `wl` and `n`.

This PR appends the fitted `wl` and `n` to the same `.txt`, so the saved file is a complete per-channel summary of the fit:

- New columns at the end: `wl[pA/rtHz]`, `n`.
- Channels whose fit was rejected (`f_knee == 0.`) get `NaN` in all three fitted columns, matching the existing `f_knees` convention — fit-success rows still have finite values everywhere.
- A `#`-prefixed header line is written so the column meaning is self-documenting and `np.loadtxt` continues to read the file cleanly.
- Existing columns 0–2 keep their order, so any consumer reading by index still works.
- Docstring for `save_data` updated to describe the five-column format.
- No API changes, no new arguments, no change to the per-channel `.npy` sidecars or to the `return_noise_params` path.

Diff is `+14 / −2` in a single file (`python/pysmurf/client/debug/smurf_noise.py`).

## Test plan

- [x] `flake8 --count python/` clean (matches CI invocation in `.github/workflows/test-or-deploy.yml`).
- [x] Offline round-trip with synthetic per-channel arrays: file has 5 columns, header is `#`-prefixed, `np.loadtxt` reads `(n_channel, 5)`, NaN preserved for rejected-fit rows, finite for successful rows.
- [ ] On-hardware smoke test: run `S.take_noise_psd(meas_time, save_data=True)` on a SMuRF carrier and verify the produced `.txt` has the new columns and reasonable values for channels with good fits.

### DIFF
--- a/python/pysmurf/client/debug/smurf_noise.py
+++ b/python/pysmurf/client/debug/smurf_noise.py
@@ -68,7 +68,11 @@ class SmurfNoiseMixin(SmurfBase):
         make_summary_plot : bool, optional, default True
             Whether to make the summary plots.
         save_data : bool, optional, default False
-            Whether to save the band averaged data as a text file.
+            Whether to save the per-channel summary as a text file. The
+            file has one row per channel and five columns:
+            ``res_freq[MHz]``, ``noise_floor[pA/rtHz]``, ``f_knee[Hz]``,
+            ``wl[pA/rtHz]``, ``n``. Channels for which the noise-model
+            fit was rejected are NaN in the fitted columns.
         show_plot : bool, optional, default False
             Show the plot on the screen.
         datafile : str or None, optional, default None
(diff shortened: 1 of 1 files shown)


ruck314:
Closing this PR because the associated issue it addresses, #186, is labelled `wontfix`. Not pursuing this change.

PR #950, opened 2026-05-01
fix(read_adc_data): Warn when ADC samples rail (#210)

### LINKED ISSUE #210: Functions that use the RF ADC data should report if the ADC is railing
People keep running into problems because the RF ADC rails.  Functions which rely on reading the 500MHz ADCs need to report when the ADC rails so users know to adjust power levels to fix.  In particular, this should be incorporated into at least the `read_adc_data` and `tune_band_serial` functions.  I think the ADC limits are -2^15, 2^15-1.

ruck314:
## Summary
- Closes #210. Adds a railing warning directly to `read_adc_data` so direct callers (notebooks, diagnostic scripts, future code paths) get told when 16-bit ADC samples are pinned at the rail.
- New `check_if_adc_is_saturated=True` kwarg on `read_adc_data` (matching the existing pattern on `full_band_resp` / PR #915's `find_freq`). When set, after the read it computes max/min over real and imag and logs a red `LOG_ERROR` if any sample is outside ±31000 (~5% of ±2**15), pointing the user at the DC attenuator. Warns rather than raises so the returned data is still available for diagnostic inspection.
- Internal callers that already do their own saturation logic — `check_adc_saturation` and the two `read_adc_data` calls inside `full_band_resp` — pass `check_if_adc_is_saturated=False` to avoid duplicate log lines. Their existing behavior (max/min log + `ValueError` raise from `full_band_resp` on rail) is preserved unchanged, so `tune_band_serial`'s railing report (which goes through `full_band_resp`) is unaffected.

## Issue scope assessment
The issue (filed 2019-12-15) called out both `read_adc_data` and `tune_band_serial`:
- **`tune_band_serial`** has been functionally addressed since commit `6e94b045` added `check_if_adc_is_saturated=True` to `full_band_resp` — it raises `ValueError` with a "Try increasing the DC attenuation" message on rail.
- **`read_adc_data`** itself was the remaining gap — it returned raw counts with no warning, even at hard ±2**15 rail. This PR closes that gap.

## Out of scope (intentionally)
- **`read_dac_data`**: issue #210 is scoped to ADC; `check_dac_saturation` already exists for DAC-side users.
- **The orphaned UC-attenuator state in `tune_band_serial` when `full_band_resp` raises** (sets `att_uc=0`, restore line is skipped on exception). Real but separate bug.
- **A firmware-side saturation register read during eta scan** (suggested by @swh76 in the issue thread). Requires Rogue/firmware changes.

## Test plan
- [x] `flake8 --count python/` → 0 errors (matches CI invocation in `.github/workflows/test-or-deploy.yml`).
- [x] Offline numpy sanity check of the threshold: clean / boundary-equal / in-range arrays correctly do not trigger; real-rails-high, imag-rails-low, just-over-31001, and hard ±2**15 cases all correctly do trigger.
- [ ] Hardware regression on a SMuRF system:
  - `S.read_adc_data(band, ...)` at deliberately too-low DC attenuation → confirm the red railing warning fires with sensible max/min counts; re-run at nominal attenuation → confirm no warning.
  - `S.check_adc_saturation(band)` → confirm only one set of log lines (no duplicated railing warning from the inner read).
  - `S.full_band_resp(band)` at saturating drive → confirm the existing `ValueError` still fires with no extra duplicate warning preceding it.
  - `S.tune_band_serial(band)` at saturating drive → confirm the `ValueError` propagates (existing behavior preserved).

### DIFF
--- a/python/pysmurf/client/tune/smurf_tune.py
+++ b/python/pysmurf/client/tune/smurf_tune.py
@@ -655,11 +655,11 @@ class SmurfTuneMixin(SmurfBase):
             # Take read the ADC data
             try:
                 adc = self.read_adc_data(band, nsamp, hw_trigger=hw_trigger,
-                    save_data=False)
+                    save_data=False, check_if_adc_is_saturated=False)
             except Exception:
                 self.log('ADC read failed. Trying one more time', self.LOG_ERROR)
                 adc = self.read_adc_data(band, nsamp, hw_trigger=hw_trigger,
-                    save_data=False)
+                    save_data=False, check_if_adc_is_saturated=False)
             time.sleep(.05)  # Need to wait, otherwise dac call interferes with adc
 
             try:
--- a/python/pysmurf/client/util/smurf_util.py
+++ b/python/pysmurf/client/util/smurf_util.py
@@ -1670,7 +1670,8 @@ class SmurfUtilMixin(SmurfBase):
            True if ADC is saturated, otherwise False.
         """
         adc = self.read_adc_data(band, data_length=2**12, make_plot=False,
-                  save_data=False, show_plot=False, save_plot=False)
+                  save_data=False, show_plot=False, save_plot=False,
+                  check_if_adc_is_saturated=False)
         adc_max   = int(np.max((adc.real.max(), adc.imag.max())))
         adc_min   = int(np.min((adc.real.min(), adc.imag.min())))
         saturated = ((adc_max > 31000) | (adc_min < -31000))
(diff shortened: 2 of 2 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. This adds new functionality to the SmurfControl mixins, which stage 6 dissolves into functions over a session, and the proposal keeps every stage net-negative in client surface. Better re-proposed on the new session API once stage 2 settles the interface. Leaving #210 open.

Issue #678, opened 2021-10-15
Ordering of commands in play_tes_bipolar_waveform


jlashner:
## Describe the bug
Currently play_tes_bipolar_waveform enables the waveform before the lut tables are set, possibly zeroing the bias lines and latching detectors.

Issue #425, opened 2020-05-01
RTM waveforms stop on whatever value they're on when asked to stop playing waveform


swh76:
## Describe the bug
Need a work around for this, otherwise when users play waveforms and then stop them they will likely be left on an indeterminate bias.  More details on the issue are logged on this JIRA ticket:
https://jira.slac.stanford.edu/browse/ESCRYODET-394

swh76:
@jmdewart I implemented this to stop TES waveforms:
https://github.com/slaclab/pysmurf/blob/fabc41fae8bf748fcd2812857b6f5d21dae0898c/python/pysmurf/client/util/smurf_util.py#L4149
is there a better way to do this such that it always stops with the starting value, or something like that?  If not, I think we'll have to cache values somewhere in pysmurf.

PR #952, opened 2026-05-01
feat(play_tes_bipolar_waveform): Add dc_amp option to play on top of existing DC bias (#184)

### LINKED ISSUE #184: Add option to play_tes_bipolar_waveform to play waveform on top of existing DC bias
play_tes_bipolar_waveform lives here:

https://github.com/slaclab/pysmurf/blob/941bddb81d87a5cf191de0112e9cf4711d7d04d5/pysmurf/util/smurf_util.py#L3312

ruck314:
## Summary

Closes #184.

Adds an optional `dc_amp` keyword argument to `play_tes_bipolar_waveform` so the supplied waveform can be played on top of an existing DC bias instead of being centered on 0 V.

- `dc_amp=None` (default) → waveform is loaded into the LUTs unmodified — existing behavior, fully backwards compatible.
- `dc_amp=<float>` → the offset is added to every sample before the LUTs are loaded.

The idiom for the issue's specific use case (play on top of the bias group's currently-programmed DC bias) is:

```python
S.play_tes_bipolar_waveform(bg, wf, dc_amp=S.get_tes_bias_bipolar(bg))
```

The naming and `None`-default convention mirrors the existing `play_sine_tes` higher-level wrapper (`smurf_util.py:3817`), which already pre-sums a `dc_amp` into a cosine waveform before calling `play_tes_bipolar_waveform`. This change pushes that pattern down to the primitive so any waveform — not just a sine wave — can be played on top of a DC offset.

This is independent of and orthogonal to the other open `play_tes_bipolar_waveform` PRs:
- #895 — enable-ordering bug (#678)
- #936 — restore pre-play bias on stop (#425)

No changes to call sites or behavior on the default code path.

## Test plan

- [x] `flake8 --count python/` reports 0 findings (matches the CI invocation in `.github/workflows/test-or-deploy.yml:51`).
- [x] Verified the only in-tree caller (`play_sine_tes` at `smurf_util.py:3864`) uses 2 positional args, so it is unaffected by the new keyword.
- [ ] Hardware smoke test on a SMuRF crate:
  ```python
  S.set_tes_bias_bipolar(bg, 1.0)
  wf = 0.1 * np.cos(2*np.pi*np.arange(2048)/2048)
  # New path: differential output should swing 0.9 V <-> 1.1 V
  S.play_tes_bipolar_waveform(bg, wf, dc_amp=S.get_tes_bias_bipolar(bg))
  S.stop_tes_bipolar_waveform(bg)
  # Regression: existing call shape still swings around 0 V
  S.play_tes_bipolar_waveform(bg, wf)
  ```

### DIFF
--- a/python/pysmurf/client/util/smurf_util.py
+++ b/python/pysmurf/client/util/smurf_util.py
@@ -4257,7 +4257,7 @@ class SmurfUtilMixin(SmurfBase):
         return hdr,row
 
     def play_tes_bipolar_waveform(self, bias_group, waveform, do_enable=False,
-            continuous=True, **kwargs):
+            continuous=True, dc_amp=None, **kwargs):
         """ Play a bipolar waveform on the bias group.
 
         Args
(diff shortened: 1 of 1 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. This adds new functionality to the SmurfControl mixins, which stage 6 dissolves into functions over a session, and the proposal keeps every stage net-negative in client surface. Better re-proposed on the new session API once stage 2 settles the interface. Leaving #184 open.

PR #958, opened 2026-05-01
feat(tools): Add fix_tmux_display helper to refresh DISPLAY in tmux (#126)

### LINKED ISSUE #126: tmux graphics always routed to display where tmux session was started
See

https://goosebearingbashshell.github.io/2017/12/07/reset-display-variable-in-tmux.html

for a possible fix.

ruck314:
## Description

Closes #126.

When a tmux session is started inside an SSH session with X11 forwarding, tmux captures the original `$DISPLAY`. After detaching and reconnecting from a different SSH session, every existing tmux pane (and every Python process already running there) keeps the stale value, so `plt.show()` opens windows on an X server that's no longer reachable.

This PR adds a small `fix_tmux_display()` helper in `python/pysmurf/client/util/tools.py` that queries `tmux show-environment DISPLAY` and writes the result into `os.environ['DISPLAY']`. Users call it after re-attaching tmux and before any interactive plotting:

```python
from pysmurf.client.util.tools import fix_tmux_display
fix_tmux_display()
```

Safe no-op outside tmux, when `tmux` is not on `PATH`, or when tmux has no `DISPLAY` recorded (`-DISPLAY` reply).

A complementary user-side remedy that doesn't require any code (and is worth recommending in docs as a follow-up) is adding `set-option -g update-environment " DISPLAY"` to `~/.tmux.conf` so newly opened panes inherit the refreshed value.

### Why a manual helper, not an auto-call from `SmurfControl.__init__`

The bug manifests *after* tmux reattach, by which point `SmurfControl` is typically already constructed. Auto-syncing only at `__init__` would not help that path, and silently rewriting `os.environ['DISPLAY']` on construction is surprising for users who deliberately point pysmurf at Xvfb or a non-default display.

## Jira Issue

ESCRYODET-

## Tests done on this branch

- `flake8 --count python/` reports 0 errors (matches the CI step in `.github/workflows/test-or-deploy.yml`).
- Smoke tests of `fix_tmux_display()` (run by directly importing `tools.py` to bypass the pyrogue import chain on the dev host):
  - Outside tmux (`TMUX` unset): returns `None`, leaves `DISPLAY` unchanged.
  - `TMUX` set but `tmux` binary missing on `PATH`: returns `None`, no exception raised.
  - `tmux show-environment DISPLAY` returns `-DISPLAY` (unset in tmux env): returns `None`, leaves `DISPLAY` unchanged.
  - `tmux show-environment DISPLAY` returns `DISPLAY=localhost:99.0`: updates `os.environ['DISPLAY']` and returns `'localhost:99.0'`.
- The actual reattach-with-stale-display scenario needs a real tmux + SSH X11 setup; that end-to-end check is left for reviewers/users on a SMuRF host.

New public function `pysmurf.client.util.tools.fix_tmux_display()`. No existing interfaces are modified.

`pysmurf.client.util.tools` had no DISPLAY/tmux helper.

```python
def fix_tmux_display() -> str | None:
    """Sync DISPLAY from the current tmux session.

    Returns the new DISPLAY value if updated, otherwise None.
    No-op outside tmux, when tmux is not on PATH, or when tmux has
    no DISPLAY recorded.
    """
```

### DIFF
--- a/python/pysmurf/client/util/tools.py
+++ b/python/pysmurf/client/util/tools.py
@@ -14,6 +14,7 @@
 # contained in the LICENSE.txt file.
 #-----------------------------------------------------------------------------
 import os
+import subprocess
 
 import numpy as np
 from scipy.optimize import curve_fit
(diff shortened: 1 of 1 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. A tmux DISPLAY helper is a developer-environment convenience, not readout code, and the proposal shrinks the client surface rather than growing it. Leaving #126 open.

Issue #114, opened 2019-10-10
Add docstrings for the UC and DC attenuator set/get functions


swh76:
The current docstrings are not helpful;

https://github.com/slaclab/pysmurf/blob/b397b6d7382fff7794a367225ed80b6a319381f6/pysmurf/command/smurf_command.py#L1717

PR #962, opened 2026-05-01
Document and warn on amplitudeScale; add dB helpers (closes #123, #114)

### LINKED ISSUE #123: Overhaul amplitudeScale and UC/DC attenauator set/get functions

ruck314:
## Summary

Closes #123 and the linked #114.

Issue #123 has been open since 2019 and asks for three things on the `amplitudeScale` and UC/DC attenuator interfaces:

1. **Better docstrings** explaining what `amplitudeScale` is and why the conventional value is 12.
2. **A warning** when `amplitudeScale != 12` (in particular `> 12`, which has empirically caused DAC railing on >10 tones at amplitudeScale 14/15).
3. **Set attenuators and tone power in dB** rather than raw register codes (Dan's suggestion).

#114 is the docstring-only subset for UC/DC attenuators and is closed by the same docstring commit.

## What changed

Four atomic commits, all additive — no API breaks, no schema changes, no changes to tuning workflows.

### `docs(amplitudeScale,att_uc,att_dc): Document units, ranges, and DAC-rail risk (#123, #114)`

Fill in real docstrings for:

- `set_amplitude_scales` / `get_amplitude_scales`
- `set_amplitude_scale_array` / `get_amplitude_scale_array`
- `set_amplitude_scale_channel` / `get_amplitude_scale_channel`
- `set_amplitude_scale_array_currentchans`
- `set_att_uc` / `get_att_uc`
- `set_att_dc` / `get_att_dc`

Replace the placeholder `amplitude_scale` property docstring (literal `?` placeholders) in `SmurfConfigPropertiesMixin` with one that documents the [0, 15] range, the 3 dB / LSB convention, the meaning of 12 = full scale with ~3 dB DAC headroom, and links to the firmware write paths and the new dB helpers.

Attenuator docstrings document the [0, 31] 5-bit range, 0.5 dB / LSB convention, and that 0 = no attenuation.

### `feat(set_amplitude_scale_array,_channel): Warn when amplitudeScale > 12 (#123)`

`set_amplitude_scale_array` and `set_amplitude_scale_channel` now log a single `WARNING:` line on the firmware write path when any element / value exceeds 12.  Values ≤ 12 stay silent so routine low-power tuning is not noisy.  No API change; existing callers behave identically except for the new log line.

### `feat(SmurfCommandMixin): Add tone_power and att dB conversion helpers (#123)`

Four new `@staticmethod` helpers on `SmurfCommandMixin`:

- `tone_power_to_db(val)` — raw amplitudeScale → dB relative to 12.  3 dB / LSB.
- `tone_power_from_db(db)` — inverse, with bounds-check ValueError outside [0, 15].
- `att_to_db(val)` — raw 5-bit attenuator code → dB.  0.5 dB / LSB.
- `att_from_db(db)` — inverse, with bounds-check ValueError outside [0, 31].

Pure conversions, no register I/O.  Existing call sites are untouched; users who want to specify in dB call e.g. `S.set_att_uc(b, S.att_from_db(8))`.

### `fix(set_amplitude_scale_array_currentchans): Pass band, not self (#123)`

Latent bug spotted while editing the file: line 1471 called `self.set_amplitude_scale_array(self, new_amp, ...)` — passing the `SmurfControl` instance as the `band` argument.  Replaced `self` with `band`.

- [ ] Yes
- [X] No

All four commits are additive: new docstrings, a new log line on >12, four new public static helpers, and a one-character bugfix in a function that was broken anyway.  No existing callers change behavior except that they now see a warning when they pass amplitudeScale > 12, which is the user-visible intent of issue #123.

## Test plan

- [ ] `flake8 --count python/` — verified clean locally after every commit (matches the only gating CI check).
- [ ] Inspect rendered Sphinx docs for the updated docstrings.
- [ ] On hardware: confirm the `WARNING:` line appears in the log when calling `set_amplitude_scale_channel(b, c, 13)` and does not appear for `val=12`.
- [ ] On hardware: confirm `S.set_att_uc(b, S.att_from_db(8))` writes raw code 16, and `S.att_to_db(S.get_att_uc(b))` round-trips.
- [ ] Confirm no behavioral regression in `find_freq` / `setup_notches` / `tune_band_serial` / `full_band_resp` (the workflows that pull `tone_power` from `self._amplitude_scale[band]`).

## Intentionally NOT changed

- No new `tone_power_db` cfg field — adding a parallel knob doubles surface area and creates ambiguity if both are set.  Defer until concrete demand.
- No `set_*_db` / `get_*_db` wrapper methods — the static helpers are sufficient and avoid maintaining two parallel paths to the same register.
- No change to the schema range.  Issue #123 wants warnings, not a hard cap.
- No changes to `find_freq`, `setup_notches`, `tune_band_serial`, or `full_band_resp`.

### DIFF
--- a/python/pysmurf/client/base/smurf_config_properties.py
+++ b/python/pysmurf/client/base/smurf_config_properties.py
@@ -2020,17 +2020,29 @@ class SmurfConfigPropertiesMixin:
     # Getter
     @property
     def amplitude_scale(self):
-        """Short description.
+        """Default per-band tone amplitude scale.
 
-        Gets or sets ?.
-        Units are ?.
+        Gets or sets the default tone amplitude scale used when a tuning
+        routine (:func:`find_freq`, :func:`setup_notches`,
+        :func:`tune_band_serial`, :func:`full_band_resp`, ...) is called
+        without an explicit ``tone_power``.  Indexed by band number, with
+        each entry a 4-bit unsigned integer in [0, 15].
+
+        Units are raw amplitudeScale codes with ~3 dB per LSB.  The
+        conventional full-scale value is 12, which leaves ~3 dB of DAC
+        headroom for the worst-case tone sum.  Values above 12 risk
+        railing the DAC, especially with many simultaneous tones.
 
         Specified in the pysmurf configuration file as
-        `amplitude_scale`.
+        ``amplitude_scale``, per band, under each ``band_N`` block.  The
+        config schema enforces the [0, 15] range; the firmware write
+        paths additionally log a warning when the value exceeds 12.
 
         See Also
         --------
-        ?
+        :func:`SmurfCommandMixin.set_amplitude_scale_array`
+        :func:`SmurfCommandMixin.set_amplitude_scale_channel`
+        :func:`SmurfCommandMixin.tone_power_to_db`
 
         """
         return self._amplitude_scale
--- a/python/pysmurf/client/command/smurf_command.py
+++ b/python/pysmurf/client/command/smurf_command.py
@@ -1468,9 +1468,14 @@ class SmurfCommandMixin(SmurfBase):
             Per-channel tone amplitudes in [0, 15], length
             ``get_number_channels(band)``.
         """
+        arr = np.array(val).astype(np.uint)
+        if np.any(arr > 12):
+            self.log(
+                f'WARNING: amplitudeScale max={int(arr.max())} > 12 on '
+                f'band {band}.  DAC may rail with many simultaneous tones.')
         self._caput(
             self._cryo_root(band) + self._amplitude_scale_array_reg,
-            np.array(val).astype(np.uint), **kwargs)
+            arr, **kwargs)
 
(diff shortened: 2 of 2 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. Stage 3 replaces the hand-written smurf_command/base_class register layer (and its docstrings) with runtime introspection and a generated shim, so this change would be discarded. Leaving #123 open to be revisited on the new layer.

PR #963, opened 2026-05-01
feat(estimate_lms_freq): Add diagnostics and multi-population detection (#96)

### LINKED ISSUE #96: Add comments and diagnostic/statistics plots to lms_freq estimator(s)
Our lms_freq_hz estimator is opaque and often fails.  It also has almost no comments.  

https://github.com/slaclab/pysmurf/blob/0624cdacb62a5fc71c8eac33acb3b5248e5ce59b/pysmurf/tune/smurf_tune.py#L3578

One thing it'd be useful for the routine to report is whether or not there are multiple populations of lms_freq_hz.

ruck314:
## Summary

Closes #96.  The `lms_freq_hz` estimator was opaque (one scalar return, almost no comments) and silent when channels locked onto a half-period or other harmonic correlation peak instead of the fundamental.

- **`flux_mod2`** — inline comments explaining cycle-padding, the z-scored template, the tiled self-correlation, the threshold-block period pick, and the parabolic peak refinement.  New `return_peaks=False` kwarg (default preserves the legacy scalar return) lets callers recover the per-channel phi0 period array and the flux-ramp cycle length.
- **`estimate_lms_freq`** — always logs `n_chan / median / mean / std` of per-channel `lms_freq_hz` and emits a `LOG_USER` warning when `scipy.signal.find_peaks` on the smoothed histogram detects more than one population.  With `make_plot=True`, writes a 2-panel diagnostic (histogram + per-channel scatter, both with median and any secondary peak markers) to `plot_dir` via the existing `register_file` pattern.  New `save_plot` / `show_plot` kwargs match the rest of the file.
- **Backward compatibility** — `estimate_flux_ramp_amp` and any external `flux_mod2` callers using the default 1-arg form keep their current scalar return.

## Test plan

- [x] `flake8 --count python/` (CI's lint job) — clean
- [x] AST-parse both new signatures
- [x] Offline smoke test of the multi-population branch on synthetic uni / bi / trimodal arrays — `find_peaks` returns 1 / 2 / 3 peaks as expected; degenerate cases (`n=3`, all-equal) report 1 peak
- [ ] Hardware: run `tracking_setup(band, meas_lms_freq=True, make_plot=True, save_plot=True)` on a SMuRF, confirm `<timestamp>_b<band>_estimate_lms_freq.png` lands in `plot_dir` and that the logged median matches the value `tracking_setup` proceeds to use
- [ ] Hardware: confirm the multi-population warning fires on a known-misbehaving band

### DIFF
--- a/python/pysmurf/client/tune/smurf_tune.py
+++ b/python/pysmurf/client/tune/smurf_tune.py
@@ -4472,11 +4472,19 @@ class SmurfTuneMixin(SmurfBase):
     def estimate_lms_freq(self, band, reset_rate_khz,
                           fraction_full_scale=None,
                           channel=None,
-                          make_plot=False):
+                          make_plot=False, save_plot=True,
+                          show_plot=False):
         """
         Attempts to estimate the carrier (phi0) rate for all channels
         on in the requested 500 MHz band (0..7) using the flux_mod2
         routine.
+
+        Always logs band-level summary statistics (channel count,
+        median / mean / std of per-channel lms_freq_hz) and a warning
+        when the per-channel distribution is multi-modal -- e.g. a
+        sub-population locked on a half- or double-period correlation
+        peak.  When `make_plot=True`, also writes a 2-panel diagnostic
+        (histogram + per-channel scatter) to `self.plot_dir`.
 
         Args
         ----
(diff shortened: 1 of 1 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. This adds new functionality to the SmurfControl mixins, which stage 6 dissolves into functions over a session, and the proposal keeps every stage net-negative in client surface. Better re-proposed on the new session API once stage 2 settles the interface. Leaving #96 open.

PR #969, opened 2026-05-01
feat(smurf_config): separate Rcable + cryostat-card cfg files (#86)

### LINKED ISSUE #86: Need a separate Rcable for TES biasing with pysmurf
And should have cryostat-card specific configuration files that encode the resistances on the cryostat card.  That way users don't have to think about how the cryostat card was loaded.

ruck314:
## Summary

Closes #86. Splits TES bias-line resistance into two new optional cfg keys so a single cryostat-card cfg file can be shared across experiment cfgs that use the same physical card:

- `Rcable` — round-trip cable + cold-resistor resistance (cryostat-card resistance excluded).
- `cryostat_card_config_file` — path to a card-specific cfg file holding `R_cryostat_card`, `high_low_current_ratio`, `pic_to_bias_group`, and `bias_group_to_pair`.

When both are provided, the loader sets `bias_line_resistance = Rcable + R_cryostat_card`. An explicit `bias_line_resistance` in the main cfg always wins, so every existing cfg loads unchanged (full backward compat).

The merge pass runs in `SmurfConfig.read()` before schema validation. If a card-cfg key is also set in the main cfg, the main cfg wins and a notice is printed — making it easy to start from a card cfg and tweak one value per deployment.

Two example card cfgs ship under `cfg_files/cryostat_cards/`:
- `cc02-06.cfg` — seeded from the existing UCSD C2-06 experiment cfg (16505 Ω card-only, 8.084 high/low ratio).
- `cc02-template.cfg` — generic C02 template matching `cfg_files/template/template.cfg` values.

This addresses the explicit TODO that's been in `smurf_config.py:478-480` since the cryostat-card schema was first added.

## Why not auto-detect via `CryoCard.get_fw_version()`?

`get_fw_version()` only distinguishes the major version (C02 vs C04), not which loadout is on the card. Cards of the same type can have different on-card resistor values across deployments (e.g. C2-06 vs C2-02 — confirmed by grep across `cfg_files/`). Auto-detection would silently use the wrong values for many real deployments, so card cfgs are referenced explicitly.

## Files

- `python/pysmurf/client/base/smurf_config.py` — new `_merge_cryostat_card_config()` helper, `Rcable` + `cryostat_card_config_file` schema entries, TODO replaced.
- `python/pysmurf/client/base/smurf_config_properties.py` — additive `Rcable` property.
- `cfg_files/cryostat_cards/{README.md,cc02-06.cfg,cc02-template.cfg}` — new directory with shape docs + two seed card cfgs.
- `README.config_file.md` — new "Cryostat-card cfg files" section + updated `bias_line_resistance` description.
- `tests/client/validate_cryostat_card_config.py` — five-scenario hardware-free validator (legacy, new, override, missing, relative-path) using `importlib` to avoid the `pysmurf.client` package's eager `pyrogue` import.

## Test plan

- [x] `flake8 --count python/` → 0 errors (matches CI).
- [x] `python tests/client/validate_cryostat_card_config.py` → all 5 tests pass.
- [x] Regression sweep: load every `cfg_files/**/experiment_*.cfg` with the merge pass + schema validation. 25/38 pass on this branch — identical to unmodified `main`. The 13 pre-existing failures are unrelated (`50k_Id_offset` / `reset_rate_khz` schema strictness on legacy cfgs).
- [ ] Real-hardware smoke: load a C2-06 deployment cfg migrated to use `cryostat_card_config_file: ../cryostat_cards/cc02-06.cfg` + `Rcable: 29.1`, confirm `bias_line_resistance` matches the previous explicit value (16534.1 Ω) and IV / noise analyses behave identically.

## Out of scope

- Migrating existing `cfg_files/**/experiment_*.cfg` files to the new keys. They keep working as-is; users can opt in. Follow-up issue can track repo-wide migration.
- Adding card cfgs for every C02/C03/C04 variant. Two seed examples is enough to demonstrate the pattern; values for other variants need experimental confirmation.
- Auto-detecting card type from `CryoCard.get_fw_version()` — see rationale above.

### DIFF
--- a/README.config_file.md
+++ b/README.config_file.md
@@ -68,7 +68,45 @@ This defines which timing master to use. The available values are "ext_ref" and 
 These are other variables that are in the top level of the config table. Several of these should probably be moved, but for now just know they exist.
 
 - `R_sh` - The resistance of the shunt resistor in Ohms
-- `bias_line_resistance`- The resistance of cabling from cryocard to the TES.
+- `bias_line_resistance`- The total round-trip TES bias-line resistance in low-current mode, in Ohms.  Includes both the inline resistance on the cryostat card and the cryocable resistance.  May be supplied directly (legacy single-value path) or derived from `Rcable` plus the cryostat-card resistance loaded from a card cfg file (see [Cryostat-card cfg files](#cryostat-card-cfg-files) below).  An explicitly set `bias_line_resistance` always takes precedence.
+- `Rcable` - Optional.  Round-trip TES bias-cable resistance plus any cold resistors, in Ohms, with the cryostat-card resistance excluded.  When paired with a cryostat-card cfg that supplies `R_cryostat_card`, the loader sets `bias_line_resistance = Rcable + R_cryostat_card` automatically.  Ignored if `bias_line_resistance` is set explicitly.
+- `cryostat_card_config_file` - Optional.  Path to a separate JSON cfg file holding cryostat-card-specific values (`R_cryostat_card`, `high_low_current_ratio`, `pic_to_bias_group`, `bias_group_to_pair`).  Resolved relative to the main cfg's directory if not absolute.  See [Cryostat-card cfg files](#cryostat-card-cfg-files) below.
 - `tune_dir`- The directory where the tuning files are stored. Tuning files define the resonator frequencies and the channel assignments
 - `default_data_dir`- The directory where the output data is stored. Output data can include anything from eta-scans to tracked resonator timestreams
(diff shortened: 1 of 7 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. Stage 4 replaces smurf_config.py / smurf_config_properties.py / template.cfg and the cfg_files layout with a single inherited config schema, so this belongs in the new schema rather than in code slated for deletion. Leaving #86 open.

PR #967, opened 2026-05-01
refactor(scripts): Rename shawnhammer to smurf_startup and move to scripts/ (#85)

ruck314:
## Summary
Closes #85. The issue asked for four things:

1. Rename `shawnhammer` to `smurf_startup` and move it out of `scratch/shawn/scripts/` into a new top-level `scripts/` directory.
2. Replace hard-coded params with a configuration file.
3. Add command-line options.
4. Add validation steps during boot.

Items 2, 3, and 4 already shipped through #429, #433, #452, and #474. This PR finishes item 1.

## Changes
- `git mv scratch/shawn/scripts/shawnhammer.sh scripts/smurf_startup.sh`
- `git mv scratch/shawn/scripts/shawnhammerfunctions.sh scripts/smurf_startup_functions.sh`
- Updated the `source` line, the cfg-missing error message, and the "Done with smurf startup" log line in the renamed scripts.
- Updated `scratch/shawn/scripts/install.sh` to softlink the new files from their `scripts/` location (so `smurf_startup` lands on PATH instead of `shawnhammer`).
- Left **backward-compat symlinks** at `scratch/shawn/scripts/shawnhammer.sh` and `scratch/shawn/scripts/shawnhammerfunctions.sh` pointing at the new files, so any deployed system that still references the old paths keeps working until it migrates.

## Test plan
- [x] `bash -n scripts/smurf_startup.sh`
- [x] `bash -n scripts/smurf_startup_functions.sh`
- [x] `bash -n scratch/shawn/scripts/install.sh`
- [x] `flake8 --count python/` (matches CI)
- [x] `git log --follow scripts/smurf_startup.sh` traces back through the rename (97% copy similarity → git preserves history).
- [ ] On a smurf workstation: `cd scratch/shawn/scripts && ./install.sh` lands `smurf_startup` and `smurf_startup_functions` symlinks in `/usr/local/src/smurf-server-scripts/docker_scripts/`, then `smurf_startup -c /data/smurf_startup_cfg/smurf_startup.cfg` boots a system end-to-end.
- [ ] Confirm the legacy `scratch/shawn/scripts/shawnhammer.sh` invocation still works through the compat symlink.

## Out of scope
- #520 (pip-installable scripts under `pysmurf/bin`) — separate, broader cleanup.
- #78 (refuse to run inside an existing tmux session) — separate behavior fix.

PR #970, opened 2026-05-01
fix(shawnhammer): Refuse to run inside an existing tmux session (#78)

### LINKED ISSUE #78: shawnhammer should refuse to run if it's accidentally run inside of a tmux session
Right now, if you accidentally run shawnhammer in a tmux session, it kills itself without any obvious errors.  It'd be nice if it knew whether or not it was being run in a tmux session and then threw an error and refused to proceed if it is instead.

ruck314:
## Summary
- Closes #78. shawnhammer manages its own tmux session via `tmux kill-session` / `tmux new-session`. When the script is launched from inside that (or any) tmux session, the surrounding shell gets torn down and the script silently dies with no obvious error.
- Adds a `$TMUX` guard at the very top of `scratch/shawn/scripts/shawnhammer.sh` that bails with a clear error before any side effects.

## Test plan
- [x] `bash -n scratch/shawn/scripts/shawnhammer.sh` — syntax OK
- [x] With `TMUX` set: prints `ERROR: shawnhammer cannot be run from inside a tmux session.` and exits 1 (no tmux session is killed)
- [x] With `TMUX` unset: guard is skipped; script proceeds to the existing arg/cfg validation (no false positive)
- [ ] Smoke-test on a real SMuRF server: `tmux new -s test78` then run `shawnhammer` — confirm it refuses; detach and re-run from a normal shell — confirm normal startup is unchanged.

## Note on PR #967
PR #967 renames this file to `scripts/smurf_startup.sh`. Whichever lands second is a trivial rebase to move the same 10-line guard into the renamed file.

### DIFF
--- a/scratch/shawn/scripts/shawnhammer.sh
+++ b/scratch/shawn/scripts/shawnhammer.sh
@@ -1,5 +1,15 @@
 #!/bin/bash
 
+# Refuse to run from inside an existing tmux session.  This script kills and
+# recreates its own tmux session below; if it is launched from inside that
+# (or any) tmux session, the surrounding shell gets torn down and the script
+# silently dies with no obvious error.  See pysmurf issue #78.
+if [ -n "$TMUX" ]; then
+    echo "ERROR: shawnhammer cannot be run from inside a tmux session." 1>&2
+    echo "       Detach from tmux first (Ctrl-b d), then re-run from a regular shell." 1>&2
+    exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. Stage 5 moves setup() server-side and replaces the shell startup scripts with a startup API, so this belongs in the server-side procedure. Leaving #78 open.

PR #971, opened 2026-05-01
feat(SmurfConfig): Re-add tone/analysis/synthesisScale as optional band cfg overrides (#67)

### LINKED ISSUE #67: Add synthesisScale, toneScale, and analysisScale back to pysmurf cfg in the dspv3 branch as optional params
I took these out of dspv3 with the thought that these should only be set from the defaults.yml, but it's useful to be able to override on occasion.

ruck314:
## Summary

Closes #67. Restores the per-band override path for the three DSP scale registers — `toneScale`, `analysisScale`, and `synthesisScale` — that were stripped from the cfg flow in commit 266b58c3 (May 2019) on the assumption that `defaults.yml` would be the only place they were set.

The override stays **opt-in**: cfgs that omit these keys behave exactly as before (whatever `defaults.yml` programmed stays in place). Cfgs that include them now have those values written to firmware during `setup()`, which is the behavior the issue asks for.

## Changes

- `python/pysmurf/client/base/smurf_config.py` — Add 3 `Optional(default=None)` entries to the `band_#` schema with bound `0..15` (matches `amplitude_scale`).
- `python/pysmurf/client/base/smurf_config_properties.py` — Add `_tone_scale` / `_analysis_scale` / `_synthesis_scale` private fields, dict-comprehension loaders in `copy_config_to_properties`, and `@property`/`@setter` accessors. Loaders use `.get()` so missing keys flow through as `None`.
- `python/pysmurf/client/base/smurf_control.py` — Three None-guarded setter calls in the `setup()` per-band loop, just after `set_lms_gain`. Pattern mirrors `lms_delay` (the closest existing analog).
- `README.config_file.md` — One bullet per param in the `band_#` section, marked *(optional)*.

`template.cfg` deliberately stays untouched — the issue framing is "useful to be able to override on occasion", not "every user should set these."

The setter/getter methods themselves (`set_tone_scale` etc. in `smurf_command.py`) were never removed and are reused as-is.

## Test plan

- [x] `flake8 --count python/` (matches CI in `.github/workflows/test-or-deploy.yml`) — 0 errors.
- [x] Schema regression — `cfg_files/template/template.cfg` validates and exposes all three scales as `None` per band.
- [x] Schema acceptance — synthetic cfg with `toneScale: 5, analysisScale: 7, synthesisScale: 4` in `band_0` and no scales in `band_1` validates; values flow through to `band_0`, `None` for `band_1`.
- [x] Schema rejection — `toneScale: 16` is rejected by `SchemaError` as expected.
- [x] Property plumbing — `__init__` defaults the three private fields to `None`, `@property`/`@setter` round-trip works, dict-comprehensions handle full / partial / absent overrides.
- [ ] On a real SMuRF: boot with a cfg that omits the three keys; confirm `S.get_tone_scale(band)` / `get_analysis_scale` / `get_synthesis_scale` retain the `defaults.yml` values after `setup()`. Then boot with a cfg that overrides them; confirm the registers reflect the cfg values.

### DIFF
--- a/README.config_file.md
+++ b/README.config_file.md
@@ -25,6 +25,9 @@ You should be able to call this config file when you instantiate your `SmurfCont
 - `att_uc` - The upconverter (DAC) attenuator
 - `att_dc` - The downconverter (ADC) attenuator
 - `amplitude_scale` - The probe tone power.
+- `toneScale` *(optional)* - Override of the DSP `toneScale` register. Omit to keep the value programmed by `defaults.yml`.
+- `analysisScale` *(optional)* - Override of the DSP `analysisScale` register. Omit to keep the value programmed by `defaults.yml`.
+- `synthesisScale` *(optional)* - Override of the DSP `synthesisScale` register. Omit to keep the value programmed by `defaults.yml`.
 
 ## bad_mask
 
--- a/python/pysmurf/client/base/smurf_config.py
+++ b/python/pysmurf/client/base/smurf_config.py
@@ -459,6 +459,17 @@ class SmurfConfig:
 
                 # LMS gain, powers of 2
                 'lmsGain': And(int, lambda n: 0 <= n < 2**3),
+
+                # Optional DSP scale overrides.  When omitted, pysmurf
+                # does not touch the firmware register, so whatever
+                # value defaults.yml programmed stays in effect.  See
+                # issue #67.
+                Optional('toneScale', default=None):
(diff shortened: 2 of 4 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. Stage 4 replaces smurf_config.py / smurf_config_properties.py / template.cfg and the cfg_files layout with a single inherited config schema, so this belongs in the new schema rather than in code slated for deletion. Leaving #67 open.

PR #972, opened 2026-05-01
feat(smurf_util): Add set_fixed_tones for fast batch fixed-tone placement (#66)

### LINKED ISSUE #66: Need faster routines for setting arrays of fixed tones
The `set_fixed_tone` function described in https://github.com/slaclab/pysmurf/issues/58 can be used to set fixed tones, but it's slow.  Should make a version that takes a list of frequencies to set fixed tones at which uses the much faster array sets/gets.
In the meantime, here's an example of how to make combs of fixed tones in my scratch using the existing (slow) routines:
https://github.com/slaclab/pysmurf/blob/dspv3/scratch/shawn/make_fixed_tone_comb.py

ruck314:
## Summary

Closes #66 by adding a plural `set_fixed_tones(freqs_mhz, tone_power)` method in `python/pysmurf/client/util/smurf_util.py`. It places every requested frequency using the bulk array set/get registers (`set_center_frequency_array`, `set_amplitude_scale_array`, `set_feedback_enable_array`) instead of per-channel writes.

For *N* tones spanning *M* bands the call issues exactly **3·M** EPICS array writes (vs **3·N** per-channel writes for a `set_fixed_tone` loop). The 256-tone comb in `scratch/shawn/make_fixed_tone_comb.py` drops from 768 caputs to 3 — the speedup users asked for in #66.

The scalar `set_fixed_tone` is unchanged; existing callers (`smurf_noise.py:1988` and several scratch scripts) keep working.

## Behaviour

- Each frequency is placed on the band whose center is closest, using the lowest unallocated channel in its subband — same selection rule as the scalar.
- Intra-call allocations are tracked, so two requests in the same subband can't pick the same channel.
- Failure semantics match the scalar: asserts on out-of-band freq or full subband **before** any caputs, so hardware is untouched on error.
- Returns parallel `(bands, channels)` numpy arrays describing where each input frequency was placed.
- Removes the stale `# SHOULD MAKE A GET FIXED TONE CHANNELS FUNCTION` TODO comment that this function fulfils.

## Prior art

PR #796 (`feat(smurf_util.turn_on_fixed_tones)`) was an earlier attempt that was closed without merging — it had a bug (passed `amp` instead of `feedback` to `set_feedback_enable_array`) and only re-enabled tones on existing channels by index, so it didn't address the "list of frequencies" request in #66.

## Test plan

- [x] `flake8 --count python/` — returns 0
- [x] `python -m py_compile python/pysmurf/client/util/smurf_util.py` — clean
- [x] AST check confirms `set_fixed_tone`, `set_fixed_tones`, and `turn_off_fixed_tones` all present with expected signatures
- [ ] Run `make_fixed_tone_comb.py` adapted to use `set_fixed_tones` on real hardware and confirm the resulting tone comb is identical to the loop-based version, with wall-clock dropping from minutes to sub-second
- [ ] Confirm assertion still fires for an out-of-band frequency in batch mode without leaving partial state on hardware

### DIFF
--- a/python/pysmurf/client/util/smurf_util.py
+++ b/python/pysmurf/client/util/smurf_util.py
@@ -4011,8 +4011,120 @@ class SmurfUtilMixin(SmurfBase):
 
         return band, channel
 
-    # SHOULD MAKE A GET FIXED TONE CHANNELS FUNCTION - WOULD MAKE IT
-    # EASIER TO CHANGE THINGS FAST USING THE ARRAY GET/SETS
+    def set_fixed_tones(self, freqs_mhz, tone_power, write_log=False):
+        """
+        Places fixed tones at every requested frequency, using the bulk
+        array set/get registers so the EPICS write count is independent
+        of the number of tones.  For ``N`` tones spanning ``M`` bands,
+        the call issues exactly ``3*M`` array writes (vs ``3*N``
+        per-channel writes for a loop over `set_fixed_tone`).
+
+        Each frequency is placed on the band whose center is closest to
+        it, using the lowest unallocated channel in that frequency's
+        subband (a channel is "allocated" if it has non-zero amplitude
+        or has been claimed earlier in this same call).  Failure
+        semantics match `set_fixed_tone`: asserts (and writes nothing)
+        if any frequency falls outside a usable 500 MHz band, or if
+        the target subband has no unallocated channels.
+
+        Args
+        ----
+        freqs_mhz : array-like of float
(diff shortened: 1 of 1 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. This adds new functionality to the SmurfControl mixins, which stage 6 dissolves into functions over a session, and the proposal keeps every stage net-negative in client surface. Better re-proposed on the new session API once stage 2 settles the interface. Leaving #66 open.

PR #974, opened 2026-05-01
feat(smurf_noise): Add estimate_ip3 two-tone IIP3 measurement (#59)

### LINKED ISSUE #59: Need IP3 measurement function
We should be able to write, using existing pysmurf functions, a tool for measuring the IP3 of whatever's connected to SMuRF (at least in units of amplitude scale, which is all SMuRF cares about anyway).  Solving https://github.com/slaclab/pysmurf/issues/58 first would help.

ruck314:
## Summary

Closes #59. Adds `SmurfNoiseMixin.estimate_ip3`, a single-power input-referred third-order intercept estimator built on the existing `set_fixed_tone` and `read_adc_data` routines (as suggested in the issue thread).

## API

```python
result = S.estimate_ip3(freq_mhz_a, freq_mhz_b, tone_power,
                        data_length=2**19, search_window_khz=10.0,
                        make_plot=True, show_plot=False, save_plot=True,
                        save_data=False, write_log=False)
```

Both fundamentals **and** both 3rd-order IM products (`2*f_lo - f_hi`, `2*f_hi - f_lo`) must lie within the same 500 MHz band; this is asserted up-front with a clear message.

## How it works

1. Plays equal-amplitude fixed tones at `f_lo, f_hi` (`set_fixed_tone` x2).
2. Captures wide-band ADC data on the relevant band (`read_adc_data`).
3. Cleans up by zeroing **only** the two channels it allocated, in a `finally` block — pre-existing fixed tones the caller had set are left untouched.
4. Computes a complex-input PSD via `scipy.signal.welch` (same call shape as the snippet inside `read_adc_data`).
5. Peak-searches `±search_window_khz` around each expected line to absorb FFT-bin quantization.
6. Computes
   ```
   IIP3 [dB amp_scale] = 20*log10(tone_power) + (mean(P_fund_dB) - mean(P_IM3_dB)) / 2
   ```
   averaging the two fundamentals and the two IM3 products.

Returns a dict with `iip3_amp_scale_db`, the four line powers, the four line frequencies, the band/channels used, and the timestamp. Optional annotated PSD plot (vertical markers at `f1`, `f2`, `2f1-f2`, `2f2-f1`) saved under `plot_dir`; optional `.npz` of `(f, p, result)` under `output_dir`. File registration via `self.pub.register_file(..., 'ip3', ...)` matches the pattern in `read_adc_data` / `take_noise_high_bandwidth`.

## Test plan

CI here is hardware-only (`test_loopback.py` against a self-hosted SMuRF rig), so this PR ships with manual verification:

- [ ] On a loopback-configured SMuRF, run e.g.
      ```python
      S.band_off(0)
      r = S.estimate_ip3(4900.5, 4900.7, tone_power=12,
                         make_plot=True, save_plot=True)
      print(r['iip3_amp_scale_db'])
      ```
      Confirm the PSD plot shows two large fundamentals and two smaller IM3 sidebands and that `r['iip3_amp_scale_db']` is well above `20*log10(12) ≈ 21.6` for a near-linear path.
- [ ] Confirm channel cleanup: `S.which_on(0)` returns to its pre-call set after `estimate_ip3` returns (and also after a deliberate failure injected mid-call).
- [ ] Pick a frequency pair that pushes one IM3 product just outside the 500 MHz band; verify the assertion fires with a useful message.
- [ ] `python -m pyflakes python/pysmurf/client/debug/smurf_noise.py` is clean.

### DIFF
--- a/python/pysmurf/client/debug/smurf_noise.py
+++ b/python/pysmurf/client/debug/smurf_noise.py
@@ -2050,6 +2050,201 @@ class SmurfNoiseMixin(SmurfBase):
         return ff, pxx
 
 
+    def estimate_ip3(self, freq_mhz_a, freq_mhz_b, tone_power,
+                     data_length=2**19, search_window_khz=10.0,
+                     make_plot=True, show_plot=False, save_plot=True,
+                     save_data=False, write_log=False):
+        """
+        Two-tone third-order intercept (IIP3) estimate, in dB
+        amplitude-scale units.
+
+        Plays equal-amplitude fixed tones at ``freq_mhz_a`` and
+        ``freq_mhz_b`` (which must fall in the same 500 MHz band),
+        captures raw ADC data, and computes IIP3 from the ratio of the
+        average fundamental power to the average 3rd-order
+        intermodulation product power:
+
+            IIP3 [dB amp_scale] = 20*log10(tone_power)
+                                  + (P_fund_dB - P_IM3_dB) / 2
+
+        The two allocated channels are returned to amplitude 0 on exit
+        (including on failure), so any pre-existing fixed tones the
+        caller had set are left untouched.
+
+        Args
+        ----
+        freq_mhz_a : float
+            First tone frequency in MHz (absolute).
+        freq_mhz_b : float
+            Second tone frequency in MHz (absolute). Must differ from
+            ``freq_mhz_a`` and lie in the same 500 MHz band.
+        tone_power : int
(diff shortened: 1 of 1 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. This adds new functionality to the SmurfControl mixins, which stage 6 dissolves into functions over a session, and the proposal keeps every stage net-negative in client surface. Better re-proposed on the new session API once stage 2 settles the interface. Leaving #59 open.

PR #976, opened 2026-05-01
feat(publisher): Register remaining file writes with self.pub (#55)

### LINKED ISSUE #55: All functions that write data files and plots should publish the path to the files
At the Berkeley SO F2F, the SO DAQ group requested that all data written to disk by pysmurf have the path of the written data files published.  This is only useful if the publisher is added to the pysmurf dspv3 branch, which is https://github.com/slaclab/pysmurf/issues/54.  The dspv3 branch is the main development branch at the moment.  
  Right now, files are written in a bunch of different ways throughout pysmurf - sometimes as text files, sometimes as npy files, etc..  See e.g. 

https://github.com/slaclab/pysmurf/blob/be01b4e61abfda5e4ef77f60a8bfe6a05d2f56b2/tune/smurf_tune.py#L2869

It'd be nice to do a survey of all the kinds of data files we're currently writing and add wrapper functions to `pysmurf.util` with options like publish=True (by default) so all data files of a given type are written through the same interface, and in the same way.

ruck314:
## Summary

Closes the spirit of #55 by adding `self.pub.register_file()` calls to the 15 remaining unpublished `plt.savefig` / `np.save` sites in `tune/`, `debug/`, and `util/` so the SO DAQ archiver can pick up every output file.

The publisher infrastructure that #55 asked for is already in place on `main` — `Publisher.register_file(...)` lives at `python/pysmurf/client/util/pub.py:186`, attached to every `SmurfControl` instance as `self.pub`, and the specific call site #55 called out (`tune/smurf_tune.py:2869` in `be01b4e6`) is now the `full_band_resp` block at smurf_tune.py:784–794 and publishes all three of its `.txt` outputs. The maintainers chose the scattered-`register_file` pattern over the centralized `pysmurf.util` wrapper-functions originally suggested in the issue body, and that pattern is now used in dozens of sites across the client. This PR brings the remaining stragglers in line.

## Changes

- **`python/pysmurf/client/tune/smurf_tune.py`** — `analyze_IQ_stream` (`iq_svd`) and `optimize_lms_delay` plots.
- **`python/pysmurf/client/debug/smurf_noise.py`** — `wl_list` / `f_knee_list` / `n_list` `.npy` saves; per-channel `noise_vs_bias` plot; NEI summary histogram; NEP summary histogram; per-channel `noise_vs_tone` plot; `take_noise_high_bandwidth` single-channel plot; `take_noise_high_bandwidth` PSD `.npy` save; `plot_svd_summary`; `plot_svd_modes`.
- **`python/pysmurf/client/debug/smurf_iv.py`** — `estimate_phase_transition` raw-bias plot (`iv_estimate_bias`).
- **`python/pysmurf/client/util/smurf_util.py`** — `identify_probe_tones` gap plot (`probe_tone_gap`).

Coverage of file-write sites in those four files: **39/55 → 54/55** (70.9% → 98.2%).

## Out of scope (separate follow-up)

`debug/smurf_noise.py:2047–2048` writes both the frequency array `ff` and the PSD array `pxx` to the same path `f'{filename}'`, so `ff` is silently overwritten by `pxx`. This PR leaves the bug alone (surgical changes only) and registers only the surviving file (`pxx`). I'll file a separate issue for the collision.

## Test plan

- [x] `flake8 --count python/` (matches `.github/workflows/test-or-deploy.yml:51`) — 0 errors.
- [x] `git diff` — every changed line is either a `path = ...` hoist of an existing `os.path.join(...)`, the unchanged `savefig`/`save` call rewritten to take the local `path`, or a new `self.pub.register_file(...)` line. No behavior changes outside the publisher.
- [x] Coverage check: `grep -nE '(plt\.savefig|fig\.savefig|np\.save\b|np\.savetxt|np\.savez)' python/pysmurf/client/{tune,debug,util}/*.py` returns 55 sites; manual review shows 54 are now paired with `register_file` (the 1 remaining is the deliberate ff/pxx collision noted above).
- [ ] CI test-server / test-client / test-docs run in PR — local Docker run not available on this host; the changes are line-local and don't alter imports.

Closes #55 once merged.

### DIFF
--- a/python/pysmurf/client/debug/smurf_iv.py
+++ b/python/pysmurf/client/debug/smurf_iv.py
@@ -1317,8 +1317,10 @@ class SmurfIVMixin(SmurfBase):
                            ha='right', va='top')
 
                 plt.tight_layout()
-                plt.savefig(os.path.join(self.plot_dir, iv_raw_dat['basename'] +
-                    f'_estimate_bias_bg{bg}.png'), bbox_inches='tight')
+                path = os.path.join(self.plot_dir, iv_raw_dat['basename'] +
+                    f'_estimate_bias_bg{bg}.png')
+                plt.savefig(path, bbox_inches='tight')
+                self.pub.register_file(path, 'iv_estimate_bias', plot=True)
 
             if not show_plot:
                 plt.close(fig)
--- a/python/pysmurf/client/debug/smurf_noise.py
+++ b/python/pysmurf/client/debug/smurf_noise.py
@@ -260,12 +260,21 @@ class SmurfNoiseMixin(SmurfBase):
                 # Publish the data
                 self.pub.register_file(outfn, 'noise_timestream', format='txt')
 
-                np.save(os.path.join(self.output_dir,
-                    f'{basename}_wl_list'), wl_list)
-                np.save(os.path.join(self.output_dir,
-                    f'{basename}_f_knee_list'), f_knee_list)
-                np.save(os.path.join(self.output_dir,
-                    f'{basename}_n_list'), n_list)
+                wl_path = os.path.join(self.output_dir,
+                    f'{basename}_wl_list')
+                np.save(wl_path, wl_list)
+                self.pub.register_file(wl_path, 'noise_timestream',
+                    format='npy')
+                f_knee_path = os.path.join(self.output_dir,
+                    f'{basename}_f_knee_list')
+                np.save(f_knee_path, f_knee_list)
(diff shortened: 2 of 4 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. This adds new functionality to the SmurfControl mixins, which stage 6 dissolves into functions over a session, and the proposal keeps every stage net-negative in client surface. Better re-proposed on the new session API once stage 2 settles the interface. Leaving #55 open.

PR #978, opened 2026-05-01
feat(SmurfBase): Add batch_mode for foolproof headless plotting (closes #51)

### LINKED ISSUE #51: Need a foolproof way of disabling interactive plotting
At the Berkeley SO 2019 F2F, some users report that some functions (even functions that have a show_plots flag) plot no matter what (ie, even if show_plots=False).  We need a foolproof way to put pysmurf in "batch" mode.  Maybe an member variable like _batch_mode, set and get functions, and the ability to specify _batch_mode=True or False when the pysmurf object is initialized?

ruck314:
## Summary

Resolves [#51](https://github.com/slaclab/pysmurf/issues/51) — *"Need a foolproof way of disabling interactive plotting"*.

Adds a new `batch_mode` flag (with matching `set_batch_mode` / `get_batch_mode` accessors) on `SmurfBase`, threaded through `SmurfControl`. When enabled, two layers of defense make interactive plotting genuinely impossible:

| Layer | Mechanism | Effect |
|---|---|---|
| 1 | `plt.switch_backend('Agg')` | Non-interactive backend — `plt.show()` is a no-op, no GUI possible |
| 2 | `plt.ioff()` + `plt.close('all')` | Disables auto-display in Jupyter/IPython; clears any existing figures |

This is a deliberately surgical change (2 files, +58/-3) — none of the existing 21 `plt.show()` sites or `plt.ion/ioff` toggles need to change because the backend switch makes them moot.

`save_plot=True` continues to work for writing PNGs to disk, which addresses @hlym's follow-up comment about saving plots without displaying them.

## Usage

```python
# Constructor flag
S = pysmurf.client.SmurfControl(cfg_file=..., batch_mode=True)

# Or toggle later
S.set_batch_mode(True)
S.get_batch_mode()  # -> True
```

## Notes

- `set_batch_mode(True)` is one-way for the life of the Python process. `set_batch_mode(False)` clears the flag but does not restore a GUI backend (you must restart Python). This matches matplotlib's recommended post-import switch behavior.
- The CI workflow `.github/workflows/test-or-deploy.yml:231` currently does `matplotlib.use("Agg")` as a workaround before importing pysmurf — concrete evidence the pain point was real. That workaround continues to work; this change is additive.

## Test plan

- [x] `flake8 --count python/` passes (matches CI invocation)
- [x] Offline-mode verification: constructor flag, default, runtime toggle, bool coercion, `plt.show()` no-op, `savefig` works under Agg
- [ ] Live verification on a SMuRF system: confirm `batch_mode=True` keeps plots from popping up over X11/SSH

### DIFF
--- a/python/pysmurf/client/base/base_class.py
+++ b/python/pysmurf/client/base/base_class.py
@@ -54,6 +54,14 @@ class SmurfBase:
     script_id : str or None, optional, default None
         Script id included with publisher messages. For example,
         the script or operation name.
+    batch_mode : bool, optional, default False
+        Foolproof non-interactive plotting mode (issue #51). When
+        True, switches the matplotlib backend to ``Agg`` and turns
+        off interactive mode, so plots will never pop up regardless
+        of any per-call ``show_plot`` argument or ambient
+        ``plt.ion()`` state. ``save_plot=True`` continues to work
+        for writing PNGs to disk. Once switched to Agg the backend
+        cannot be switched back without restarting Python.
     """
 
     _base_args = ['verbose', 'logfile', 'log_timestamp', 'log_prefix',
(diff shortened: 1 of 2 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. Stage 6 dissolves the SmurfControl mixin object into functions over a session, so a constructor flag on SmurfBase would be discarded. Leaving #51 open.

PR #982, opened 2026-05-01
fix(find_freq): Exclude ADC interleaving spurs from resonance candidates (#28)

### LINKED ISSUE #28: Avoid ADC interleaving spurs when assigning tones to resonator candidates
Every band has two resonator-like features caused by interleaving spurs in the ADC.  Here's an example sweep of a 500 MHz band in our cryostat on campus in which there were no resonators:
image
The spurs are the two objects identified by this find_freq full 500 MHz band sweep as resonances (more details on this data I took here - https://confluence.slac.stanford.edu/pages/viewpage.action?pageId=249335319).

Mitch says these are "interleaving spurs" and that we can "freeze offset correction in the ADC
and they appear to go away but they don't reallyand can reappear as things drift."  Some notes from Mitch & Dan:

{noformat}
The offset corrector block nullifies dc, fS/8, fS/4, 3 fS/8, and fS/2

The corrector continuously processes the data coming from the interleaving ADC cores and cannot distinguish if the tone at these frequencies is part of signal or if the tone originated from a mismatch among the interleaving ADC core
{noformat}

They look a lot like resonances in the etaScans too, so we need to avoid assigning channels to these.  Here's what the eta scans for these two look like:

image
image

ruck314:
## Summary

Resolves [#28](https://github.com/slaclab/pysmurf/issues/28) — every band has two ADC interleaving spurs at **-134.4 MHz** and **+172.8 MHz** (relative to band center) that look like resonances and were being assigned channels by `find_freq()`.

The fix filters these candidates at the single source of truth — inside `find_freq()` immediately after `find_all_peak()` returns and before storage. All downstream consumers (`setup_notches`, `assign_channels`, the saved `_resonance.txt` file) automatically inherit the cleaned list.

## Changes (single file: `python/pysmurf/client/tune/smurf_tune.py`)

- **`SmurfTuneMixin` class constants**:
  - `ADC_SPURS_MHZ = (-134.4, 172.8)` — spur centers in MHz, relative to band center
  - `ADC_SPUR_EXCLUSION_MHZ = 0.2` — ±200 kHz half-width (2× the ~100 kHz observed in @jmdewart's eta-scan plots, to absorb the drift Mitch flagged)
- **New private helper `_exclude_adc_spurs(res_freq, spurs_mhz, exclusion_mhz, band)`** — drops candidates within the exclusion window, logs `WARNING:` per drop plus a summary line. No-ops on empty input or empty spur list.
- **`find_freq()`** gains two new kwargs `adc_spurs_mhz=None` / `adc_spur_exclusion_mhz=None` (sentinel-`None` resolves to the class default; pass `[]` to disable the filter for diagnostic use). Calls the helper between `find_all_peak()` and the `freq_resp` store.
- Docstring updated in pysmurf style.

## What is NOT touched

- `set_center_frequency_array` (`smurf_command.py:2054`) — adding a warning there would be very chatty and can't cleanly identify spur proximity at the subband-relative-offset layer; filtering upstream catches every code path that flows into it.
- `assign_channels`, `setup_notches`, `find_all_peak`, `find_peak` — generic infrastructure, not the place for hardware-specific knowledge. They inherit the filtered list automatically.

## Test plan

- [x] `flake8 --count python/` reports `0` violations (matches CI lint gate).
- [x] Helper logic exercised offline against a hand-crafted array `[-200, -134.45, -134.5, 0, 50, 172.75, 172.95, 200]` MHz with default `(-134.4, 172.8)` ±0.2 — drops the 4 spur entries, keeps the 4 real ones.
- [x] Opt-out path (`adc_spurs_mhz=[]`) returns input unchanged and logs `LOG_INFO`.
- [x] Empty input array returns empty.
- [ ] **Hardware run**: invoke `find_freq(band=N)` on a band with previously-visible spurs; confirm two `WARNING:` log lines near -134.4 and +172.8 MHz plus a summary line, and that `<timestamp>_b<N>_amp_sweep_resonance.txt` no longer contains entries in `[-134.6, -134.2]` or `[172.6, 173.0]` MHz.
- [ ] **Hardware run**: `setup_notches(band=N)` immediately afterward; confirm `get_center_frequency_array(N)` no longer shows entries near `band_center - 134.4` or `band_center + 172.8`.
- [ ] **Override check**: `find_freq(band=N, adc_spurs_mhz=[])` brings back the spur candidates (proves opt-out works on real hardware).

### DIFF
--- a/python/pysmurf/client/tune/smurf_tune.py
+++ b/python/pysmurf/client/tune/smurf_tune.py
@@ -35,6 +35,13 @@ class SmurfTuneMixin(SmurfBase):
     """
     This contains all the tuning scripts
     """
+    # ADC interleaving spur centers in MHz, relative to band center.
+    # Each band has two spurs that look like resonances; find_freq()
+    # filters resonance candidates within +/- ADC_SPUR_EXCLUSION_MHZ
+    # of any spur. See issue slaclab/pysmurf#28.
+    ADC_SPURS_MHZ = (-134.4, 172.8)
+    ADC_SPUR_EXCLUSION_MHZ = 0.2
+
     @set_action()
     def tune(self, load_tune=True, tune_file=None, last_tune=False,
              retune=False, f_min=.02, f_max=.3, df_max=.03,
@@ -3440,7 +3447,8 @@ class SmurfTuneMixin(SmurfBase):
             plotname_append='', window=50, rolling_med=True,
             make_subband_plot=False, show_plot=False, grad_cut=.05,
             flip_phase=False, grad_kernel_width=8,
-            amp_cut=.25, pad=2, min_gap=2):
+            amp_cut=.25, pad=2, min_gap=2,
+            adc_spurs_mhz=None, adc_spur_exclusion_mhz=None):
         '''
         Finds the resonances in a band (and specified subbands)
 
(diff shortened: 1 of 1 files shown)


ruck314:
Closing in light of Tristan's refactor proposal. This adds new functionality to the SmurfControl mixins, which stage 6 dissolves into functions over a session, and the proposal keeps every stage net-negative in client surface. Better re-proposed on the new session API once stage 2 settles the interface. Leaving #28 open.

PR #983, opened 2026-05-01
fix(JesdWatchdog): Extend to monitor both bays (#20)

### LINKED ISSUE #20: JesdWatchdog needs to be extended to bay 1
Right now, only runs on bay 0.  However, we haven't had much need for JesdWatchdog lately...

ruck314:
## Summary
- Closes #20.
- `JesdWatchdog` previously hardcoded the bay index to `0` in every PV path (`AppTopJesd[0]`, `MicrowaveMuxCore[0]`), so a JESD drop on bay 1 was silently ignored on dual-AMC systems.
- Watchdog now subscribes per-bay `Rx`/`Tx` `DataValid` PVs; the valid-changed callback parses the bay index out of the incoming `pvname` and dispatches the matching reset thread. `jesdRXReset` / `jesdTXReset` take a `bay` arg and apply it to both `AppTopJesd[bay]` and `MicrowaveMuxCore[bay]`.
- `start_jesd_watchdog(epics_prefix, bays=None)` now requires the EPICS prefix (the watchdog is the only EPICS-CA consumer in pysmurf — the client itself runs over ZMQ to rogue, so the prefix has to come from the caller) and defaults `bays` to `self.bays`. The previous zero-arg form silently fell through to the placeholder prefix `test_epics` inside the watchdog's `__main__`, so it never actually worked against a real system.
- Direct CLI usage stays backward-compatible: `argv[1] = prefix`, `argv[2:] = bays` (defaults to `[0]` when omitted).

## Test plan
- [x] `flake8 --count python/` clean (matches CI invocation in `.github/workflows/test-or-deploy.yml`).
- [x] `python python/pysmurf/client/watchdog/JesdWatchdog.py myprefix 0 1` starts cleanly with no exceptions.
- [x] Bay-extraction regex unit-checked against `AppTopJesd[0]` and `AppTopJesd[1]` PV name forms.
- [ ] Hardware verification on a 2-bay system: force a `JesdRx:Enable` toggle on bay 1, confirm `/tmp/JesdWatchdog.log` shows `JesdRx went down on bay=1` and the recovery sequence runs against bay-1 PVs. Repeat for bay 0 to confirm no regression.

### DIFF
--- a/python/pysmurf/client/util/smurf_util.py
+++ b/python/pysmurf/client/util/smurf_util.py
@@ -272,11 +272,25 @@ class SmurfUtilMixin(SmurfBase):
 
     # the JesdWatchdog will check if an instance of the JesdWatchdog is already
     # running and kill itself if there is
-    def start_jesd_watchdog(self):
+    def start_jesd_watchdog(self, epics_prefix, bays=None):
+        """Start the JesdWatchdog subprocess monitoring the given bays.
+
+        Args
+        ----
+        epics_prefix : str
+            EPICS prefix of the running rogue server (e.g. ``'smurf_server_s5'``).
+        bays : list of int, optional
+            Bay indices to monitor. Defaults to ``self.bays`` (the bays
+            detected at server startup).
+        """
         import pysmurf.client.watchdog.JesdWatchdog as JesdWatchdog
         import subprocess
         import sys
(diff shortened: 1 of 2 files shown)


swh76:
Jesd watchdog obsolete, removed in PR https://github.com/slaclab/pysmurf/pull/999.

PR #991, opened 2026-05-13
fix(base.base_class): Disable VirtualClient monitor thread

tristpinsm:
Newer versions of rogue have added to the stop function that was used previously, and calling it now closes the ZMQ connection. This change disables only the monitor thread explicitly.

This change is required in order to use `rogue v6.12.0`.


tristpinsm:
superseded by #1011

PR #992, opened 2026-05-14
Fix asymmetric DSP bandwidth boundary in get_processed_channels()

### LINKED ISSUE #97: Tracking setup plots flatline for all channels in subbands 51 and 75 for all 500 MHz bands
The tracked frequency in `tracking_setup` plots in DSPv3 fw is a flat line for any channels which fall in subbands 51 and 75, but the channels are actually able to track (we know this because in Frankenpolar a couple of these channels are TES-coupled and we can read out TESs on them just fine).  Additional details in this JIRA ticket - https://jira.slac.stanford.edu/browse/ESCRYODET-391.  

Mitch looked at this some yesterday and found that even though the tracking setup plots for two channels in subband 75 look like this ; 

image (1)

if you plot all tracking setup data it looks like this (with no other channels on):

```python
f, df, sync = S.tracking_setup(band=band,reset_rate_khz=4, lms_freq_hz=20129.4, fraction_full_scale=0.540, make_plot=True, show_plot=True, save_plot=True,channel=S.which_on(band),nsamp=2**18,feedback_start_frac=0.2,feedback_end_frac=0.98,lms_gain=6)
plt.plot(f)
```

image (2)

so the channels are tracking, and the data is there, we must just have the channel mapping from fw to pysmurf wrong, for at least subbands 51 and 75, somewhere.  @jmdewart wonders if part of the issue may be the mapping from the 416 DSP channels in fw to the 512 pysmurf channels.

swh76:
## Description
  
Closes https://github.com/slaclab/pysmurf/issues/97 by fixing incorrect channel mapping in `get_processed_channels()` that caused channels 23 and 24 (subbands 51 and 75) to show flat lines in `tracking_setup` and any other function that calls `take_debug_data`.
  
The firmware determines which 416 of 512 channels to process using an **asymmetric** boundary condition from the DSP clock bandwidth:

-Fdsp/2 <= freq < Fdsp/2
  
The old code used a **symmetric** frequency cut (remove 48 from each end of the sorted frequency list), which is equivalent to `|freq| < Fdsp/2`. This excludes channel 25 (freq = -249.6 MHz, at the lower boundary) and includes channel 22 (freq = +249.6 MHz, at the upper boundary) — the opposite of what the firmware does.
  
The new code selects the 416 channels nearest to DC, with the tie-breaking rule that at equal |freq|, the negative frequency is included and the positive is excluded. This matches the firmware without hardcoding the DSP bandwidth ratio.
  
### Impact on `take_debug_data` without the fix

`decode_data()` maps the 416-sample debug stream back to 512 channel indices using `get_processed_channels()`. The one-channel discrepancy between pysmurf and firmware creates a 3-position shift in the debug stream mapping:

  | Stream position | Firmware outputs | pysmurf assigns to |
  |---|---|---|
  | 18 | ch 23 (-57.6 MHz) | ch 22 (+249.6 MHz) |
  | 19 | ch 24 (+57.6 MHz) | ch 23 (-57.6 MHz) |
  | 20 | ch 25 (-249.6 MHz) | ch 24 (+57.6 MHz) |

Result: channel 23's data appears at channel 22, channel 24's data appears at channel 23, and channel 24 displays data from the band-edge channel 25 (flat line). Only `take_debug_data` (and callers like `tracking_setup`) are affected — the streaming readout path is not impacted.

## Tests done on this branch

- Verified theoretically that the new processed channel set matches the firmware's asymmetric validity condition exactly (416 channels, correct boundary handling)
- Ran per-channel RF IQ probe (turning on each channel individually with `take_debug_data`) confirming firmware processes channel 25 and does not process channel 22
- Confirmed channels 23 and 24 now map to correct stream positions (18 and 19)
- Ran `tracking_setup` on band with channels in subbands 51/75 — no longer shows flat lines

Verification notebook from latest version of pysmurf ; [rogue6_issue97_postfix.ipynb](https://github.com/user-attachments/files/27749465/rogue6_issue97_postfix.ipynb)

### DIFF
--- a/cfg_files/b33/experiment_b33_rfc1-2_C05-40_s4_2xLB_backplane.cfg
+++ b/cfg_files/b33/experiment_b33_rfc1-2_C05-40_s4_2xLB_backplane.cfg
@@ -1,0 +1,566 @@
+{
+"epics_root" : "test_epics",
+"init": {
+
+	"dspEnable": 1,
+
+	"band_0" : {
+		"iq_swap_in" : 0,
+		"iq_swap_out" : 0,
+		"refPhaseDelay" : 6,
+		"refPhaseDelayFine" : 0,
+		# Matches system latency for LMS feedback (multiplex
+		# of 52x 9.6MHz ticks) Adjust to match refPhaseDelay*4
+		# (e.g. if refPhaseDelay = 6, lmsDelay = 24).  Mitch
+		# is checking this ; might not be refPhaseDelay*4 -
+		# need to revisit.
+		"lmsDelay" : 24,
+		# Adjust trigRstDly such that the ramp resets at the
+		# flux ramp glitch.  Not totally clear at the moment
+		# what that means.
+		"trigRstDly" : 15, # 0xF
+		"feedbackEnable": 1,
+		"feedbackGain" : 2048,
+		"feedbackPolarity" : 1,
+		"feedbackLimitkHz": 225,
+		"lmsGain": 1,
+		"att_uc": 12,
+		"att_dc": 0,
+		"amplitude_scale": 12
+	},
+
+	"band_1" : {
+		"iq_swap_in" : 0,
+		"iq_swap_out" : 0,
+		"refPhaseDelay" : 6,
+		"refPhaseDelayFine" : 0,
+		# Matches system latency for LMS feedback (multiplex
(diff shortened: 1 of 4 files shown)


tristpinsm (approved):
Nice! One more down

PR #1011, opened 2026-06-16
Support rogue 6.14

tristpinsm:
In rogue >=6.13, the way to specifiy the ZMQ client timeout changed.

Also ensures client stop method is called at exit.

Pending next release of rogue.


tristpinsm:
making this a draft since I should also bump the rogue versions in the dockers

PR #1018, opened 2026-07-07
Harden hardware logging against a stalled atca_monitor

ruck314:
## Problem

Running the thermal test intermittently "fails" with a distinctive signature: **no traceback, the full-band response looks fine, but data taking and the live temperature plots freeze for the rest of the run** (and if the freeze precedes the loopback step, loopback is skipped). The trigger is empirical — it only happens when an AMC card's FRU asset tag is long, e.g. `30C03A01-176`, which the fixed-width BSI/FRU field truncates to `30C03A01` (no dash). A short tag such as `A01-176` fits and the test passes.

## Root cause

The freeze is a stall in the hardware-logging thread that feeds `hwlog.dat` (what the live plots render):

- `get_hardware_log_entry()` polls every atca_monitor temperature getter in a bare `for key, value in d.items(): columns.append(str(value()))` loop — **no per-getter exception isolation**. One getter that raises (dropped atca_monitor → `ConnectionError`) kills the daemon logger thread outright.
- The `self._atca` `VirtualClient` was created **without** a timeout — the sibling server client gets `setTimeout(5000, 30000)`, the atca client did not — so `_atca_caget`'s `node.get()` can block **forever** against a wedged atca_monitor.

The true origin of the bad field is firmware/atca-monitor side (an over-length asset tag overrunning the 8-char BSI/FRU field consumed by the external `slaclab/smurf-atca-monitor` server). That is out of scope here; this PR makes pysmurf **robust to** a misbehaving atca_monitor so it can never stall or kill hardware logging.

## Changes

- **`base_class.py`** — set a 5s/30s timeout on the atca `VirtualClient`, matching the server client, so atca reads cannot block indefinitely.
- **`smurf_util.py` (`get_hardware_log_entry`)** — isolate each getter; on failure log once at `LOG_ERROR` and substitute `NaN`, so the logging thread keeps running and the plots keep updating.
- **`smurf_util.py` (`check_full_band_resp`)** — derive the AMC loading code (`A0#`) by scanning the serial for the `A0#` token instead of `amc_sn.split('-')[1]`, which raised `IndexError` or returned garbage when a long asset tag was truncated to a dash-less string. Falls back to `get_band_center_mhz` when undeterminable.

## Verification

Reproduced and validated with an offline (no HW/rogue/epics) harness that re-creates both patterns:
- Unguarded logger: `raised` (thread dies) on a dropped atca read, `hung` (thread stalls) on a wedged read.
- Guarded logger: survives with `NaN` in both cases.
- Tag parse: long/truncated → garbage/`IndexError` unguarded; safe fallback guarded.

`flake8 --count python/` passes (0 findings).

## Backport

A companion draft PR backports the hardware-logging isolation to the v4.0.0 line (`v4.0.1-dev`). v4.0.0 reads via `epics.caget` (which already carries a default timeout) and has no `check_full_band_resp`, so that backport is scoped to the logger-isolation fix.


python/pysmurf/client/util/smurf_util.py
```diff
@@ -4800,7 +4809,17 @@ def check_full_band_resp(self,
...
                 results_dict[bay]['amc_sn']=amc_sn
-                amc_type=amc_sn.split('-')[1]
+                # Pull the A0# loading code out of the serial number rather
+                # than assuming it is the second '-' delimited field.  A long
+                # asset tag gets truncated by the fixed width FRU field (e.g.
+                # '30C03A01-176' -> '30C03A01'), so amc_sn may have no '-' in
+                # the expected place; split('-')[1] would then raise or return
+                # garbage.  Fall back to amc_type=None (-> get_band_center_mhz).
+                amc_type=None
+                for tok in str(amc_sn).split('-'):
```
tristpinsm: I don't know if this succeeding is important, and returning `None` may be just fine, but it looks like that would be the result for the example in the comment. If you wanted to recover the SN in that case, a regex might be more effective: `.*(A\d{2}).*`

ruck314: Good catch, you're right, and it does matter. I checked the loop against the example in its own comment:

| `amc_sn` | old `split('-')[1]` | token loop | `re.search(r'A\d{2}')` |
|---|---|---|---|
| `C03-A01-11` | `A01` | `A01` | `A01` |
| `30C03A01-176` | `176` | `None` | `A01` |
| `30C03A01` | `IndexError` | `None` | `A01` |
| `C03-30C03A01` | `30C03A01` | `None` | `A01` |

So the token loop only fixed the crash, not the parse: in exactly the truncated case this PR is about it returned `None`, and `amc_type=None` falls through to `get_band_center_mhz`, which is the possibly-wrong value that `amc_auto_detect` exists to override. Silently degrading the band centers on the affected AMCs is worse than the `IndexError`, since the run completes with plausible but wrong `fc`.

Switched to a regex in a89b3bd. I used `re.search(r'A\d{2}', ...)` rather than `re.match(r'.*(A\d{2}).*', ...)`: the leading greedy `.*` backtracks to the *last* match, and the loading code precedes the unique id in the serial (`C03-A01-176`), so first-match is the safer anchor if a future id ever contains an `A##`-looking substring. Behavior on well-formed serials is unchanged, `None`/`''`/`'C03'`/`'None-None'` still fall back to `get_band_center_mhz`, and the `amc_type in ['A01','A02']` guard still bounds any bogus match.

tristpinsm (approved):

